PR #13, opened 2015-10-11
Make bin/* executable

andygrunwald:
Nothing special.
I just found out that the files in the bin/ folder are not marked as executable.
Now they are. :)


dicortazar:
This makes sense to me. Not waiting to main developer (@sduenas).

If not, please revert the change ;).

Thanks @andygrunwald !

PR #16, opened 2015-10-12
Limit number of return rows by one during matching operation in add command

andygrunwald:
During my work life as a programmer i learned that you should shift the work to the database as much as possible. This pull request is one of these.

If you add a unique identity and add a matcher, the process is like:
* Insert the identity
* Retrieve the new created identity
* Get all identites
* Loop over all identities
* Check for each identity if it matches with the new one and if the new one equal with the current looped identity, skip it

This PR just excludes the new created identity from the "get all identites". With this we just avoid the check if the current looped entitiy is equal the new inserted entity.
This will (maybe) a performance increase during the add / matching operation with a large dataset.


andygrunwald:
Updated PR to get running unit tests again

sduenas:
Looks good to me but I would like to address my comments before merge it.

In addition, do you mind to squash the two commits into a single one? I usually add the new tests to the same commit where I change the functionality. And please, include the tag `[api]` in the first line of the commit message.

andygrunwald:
@sduenas I created a new PR with your suggested adjustments. See https://github.com/MetricsGrimoire/sortinghat/pull/22

* Rebased with master
* Commits were squashed into one
* Suggested changes were also done

I will close this PR and we will follow in https://github.com/MetricsGrimoire/sortinghat/pull/22

PR #19, opened 2015-10-18
Support for specific error codes for Add command

andygrunwald:
All commands for sortinghat return an exit code after the operation is finished:
* 0 for success
* 1 for failure

In the linux / unix world every exit code > 0 means failure.
The problem with the current approach is: You can`t identity what kind of error happen, because every error is exit code 1.

In my current project ([andygrunwald/apihat](https://github.com/andygrunwald/apihat) - A REST(ful) API for MetricsGrimoire/sortinghat) i need to know what was going wrong. Was it a already existing entity? Or a database error?
The REST API should handle the cases differently.
Of course there is a text that can be parsed. But depending on a text is kind of difficult and dangerous.

This Pull Request offers specific error codes for the Add command.
The interface / exit codes for the command line won't be changed.
Only the call to `add` (in the Add command) returns the new error codes.

Only the Add command was handled, because i want to get your opinion and rate of accept of this PR first (instead of wasting hours over hours to implement the other commands).
If you accept this PR i will take case of the other commands in the same way.


jgbarah:
Just curious: why are you using sortinghat as a command, instead of "as a library"? Supposedly, it is designed so that its functionality can be consumed by calling functions, instead of executing the command...

andygrunwald:
@jgbarah I use sortinghat as a library.
But to reuse code as much as possible i call the `add` function of the AddCommand.
See here: https://github.com/andygrunwald/apihat/blob/master/apihat/api/identities.py#L74

I don`t run the command itself. I run the add functionality of the command.
That match my functionality to 100%.

jgbarah:
Great! Thanks for the clarification.

andygrunwald:
@sduenas If you have any kind of question, just ask. I am happy to help here :)

sduenas:
This is the idea that I had when I started with code errors but I didn't have time to add more error codes. The PR looks pretty good to me but I should change a pair if things related to `Command` class. See my comments to the code.

andygrunwald:
@sduenas I updated this PR:

* Merged with master to be up to date
* Removed `get_code_of_exception` and `get_exit_code` and returned the code of the exception itself
* Introduced `WrappedValueError` (to get code support for ValueError)
* Make use of `WrappedValueError` in add command

A few questions before i continue:

1. Is this fine for your with the `WrappedValueError`? Or is there a better way in Python?
2. This PR only covers the add command. But the other commands should be covered as well. Should we deal with it in a separate PR? Or introduce this into this one?
3. Any other tips you can provide that i will put into it?
4. Any blockers left?

sduenas:
Thanks for the PR, @andygrunwald, I've just merged it. Regarding your open questions:

1. I think that the exception is ok for now. On next iterations it should be a subclass of `BaseError` exception.
2. Separate PRs (or commits) for each command should be better for a fast review.

Thanks again!

andygrunwald:
Thanks @sduenas for merging.
1. I will change the parent class in one of the next commits
2. I created tickets (see above) to keep track of it ;) Sorry for email spamming.

PR #40, opened 2015-11-28
Add support for Docker / Dockerfile

andygrunwald:
This pull request add support for [Docker](https://www.docker.com/).
Adding support for Docker make sense in (minimum) two ways:

* Newcomers: It enables new people who are familiar with Docker to try out sortinghat really fast without installing python (+libs) and the other requirements on their system. Docker enables them a fast usable and isolated env.
* Extensions: Projects that depend on sortinghat can use this Dockerfile as a basement for their Dockerfile. One example is my project [andygrunwald/apihat](https://github.com/andygrunwald/apihat)

Docker provides a public docker registry called [Docker Hub](https://hub.docker.com/).
At this place open source projects can host the (pre build) docker images.
Next to this pull request i suggest to setup an automated build on DockerHub.
With offers the advantage that after every push the master branch is compiled into a latest docker image.
If a version of sortinghat will be tagged (and released) this version will be compiled into an image as well.

I am not able to set this build up, because i have no rights to the sortinghat repository.
Access rights + be able to setup a webhook is necessary for this.
How this can look like see [gitlab/gitlab-ce @ Docker Hub](https://hub.docker.com/r/gitlab/gitlab-ce/)

I raised this topic on the mailing list already: [[Metrics-grimoire] Docker and MetricsGrimoire](https://lists.libresoft.es/pipermail/metrics-grimoire/2015-November/002435.html).

/cc @jgbarah

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:2.7
+
+RUN pip install MYSQL-python SQLAlchemy Jinja2 python-dateutil
+
+ADD . /sortinghat
+
+WORKDIR /sortinghat
+
+RUN python setup.py install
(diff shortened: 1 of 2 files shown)


sduenas:
Great, we think it is really useful.

@acs can you review this PR, please?

andygrunwald:
@sduenas Are we good to go here?

acs:
@andygrunwald just some final comments after testing all. The idea is to have the image already created in Docker Hub so you don't need to create it locally. Not sure if maybe we can put howto create the image locally: 

    docker build -t metricsgrimoire/sortinghat ./

And when you are inside the sortinghat container the user is root. I think that this is ok: your are in a monouser environment, but you know, it is uncomfortable to execute commands as root. It sounds risky. Just to confirm this is the correct approach.

andygrunwald:
@acs: I will add a guide how to build the container locally in the README (today or tomorrow).
Because of the user: Yes and no. We are in a running isolated env. Afaik it is not possible to break out of a container with user root (of course it is a different thing if you are running a privileged container).
For this i think it is fine, because you can don`t break out.

sduenas:
Merged!

I've only squashed some commits to make the log more readable. Thanks again for the PR, @andygrunwald

andygrunwald:
Thank you @sduenas !
I will make another PR with how to build the docker file into the readme soon

andygrunwald:
@acs see https://github.com/MetricsGrimoire/sortinghat/pull/42

PR #45, opened 2016-01-01
Add support for environment variables

andygrunwald:
sortinghat depends currently on a config file in users home dir.
In terms of the 12 factor apps configuration settings should be applied as env vars as well.
See http://12factor.net/config

This pull request enables the configuration for the database via env vars.
All other configuration types (incl. default values) are still supported.

This env vars are supported:
* `SORTINGHAT_DB_USER`
* `SORTINGHAT_DB_PASSWORD`
* `SORTINGHAT_DB_DATABASE`
* `SORTINGHAT_DB_HOST`
* `SORTINGHAT_DB_PORT`

The names of the env vars follows the schema `APPNAME_SECTION_CONFIG`.

With this PR the usage of sortinghat for Docker or "cloud" envs is much easier.

See [12 Fractured Apps](https://medium.com/@kelseyhightower/12-fractured-apps-1080c73d481c#.dothyjoc9) for further read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ You are ready to use sortinghat!
 
 ## Configuration
 
-* Configure database parameters
+Set the database parameters via the `config` command:
+
 ```
   $ sortinghat config set db.host <mysql-host>
   $ sortinghat config set db.user <user>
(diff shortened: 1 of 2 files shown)


gpoo:
I like the idea and it is straightforward to implement Maybe it is something that could be set in every MetricGrimoire project. I will do a similar one for MLStats :-)

sduenas:
I like the idea. I only had one concern with regards to store passwords on environment variables but these posts solved my doubts:

* http://superuser.com/questions/708355/is-it-safe-to-store-critical-passwords-in-server-environment-variables
* http://security.stackexchange.com/questions/14000/environment-variable-accessibility-in-linux/14009

PR #74, opened 2016-09-26
Setup everything to upload first release to pypi

jgbarah:
Bumped version to 0.2.0rc1, to signal that it can now be installed via pypi.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include README.md
--- a/setup.py
+++ b/setup.py
@@ -25,32 +25,53 @@ import codecs
 import os
 import re
 
-from distutils.core import setup
-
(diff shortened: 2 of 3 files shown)


jgbarah:
From my point of view, this is ready to upload to pipy. Have a look at [the package in Pypi test](https://testpypi.python.org/pypi/sortinghat/0.2.0rc1). I've checked it up to having `sortinghat --help` working after installation via pip.

When you like it, I suggest that you change version to 0.2.0, and commit it. I can upload it then to Pypi live.

jgbarah:
I've pushed a new version of the commit, now including the MANIFEST.md file to avoid problems with the sdist package when pandoc is not available.

jgbarah:
@sduenas, anything about this?

PR #79, opened 2016-12-28
Adding dependencies, and pushing version.

jgbarah:
It seems this works with pip install now...

I ckecked with "sortinghat init test", and I got no error message (well, I got nothing at all, but I'm assuming that's ok ;-) ).


jgbarah:
Anything new about this?

PR #80, opened 2017-01-30
Add some more details about Sorting Hat in README.md

jgbarah:
Some of these details are about Sorting Hat itself, some others are about how to install it via pip.

I bumped the version label to 0.2.0 as well, to have a relatively stable release that we can upload to pypi.


jgbarah:
Done!!

PR #95, opened 2017-10-04
[cmd:load] Add option to clear ids relationships and enrollments

### LINKED ISSUE #94: Reset unique identities before loading
The command `load`appends the data from the given file to the data stored in the database. Sometimes, we need to have exactly the same relationships between identities that we have on the input file.

The idea is to have a flag named `--reset` that will remove all the identities assigned to the unique identities. Identities will be moved to a single unique identity with the same id. In other words, 1:N relationships will be removed ending up with 1:1.

Information about enrollments must be removed too.

sduenas:
Before loading any data, if 'reset' option is set, all the
relationships between identities and their enrollments will
be removed from the database.

Thus, both input and database will have the same status for
the loaded unique identities.

This commit fixes #94

### DIFF
--- a/sortinghat/cmd/load.py
+++ b/sortinghat/cmd/load.py
@@ -27,7 +27,7 @@ import sys
 
 from .. import api
 from ..command import Command, CMD_SUCCESS, HELP_LIST
-from ..db.model import MIN_PERIOD_DATE, MAX_PERIOD_DATE
+from ..db.model import MIN_PERIOD_DATE, MAX_PERIOD_DATE, Enrollment
 from ..exceptions import AlreadyExistsError, NotFoundError,\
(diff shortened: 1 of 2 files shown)


sortinghat/cmd/load.py
```diff
@@ -49,6 +49,9 @@ class Load(Command):
     engine is selected. The parameter '--match-new' can also be used to match
     only those identities that are new on the registry.
 
+    Previous relationships between identities and their enrollments will be
```
acs: «Previous relationships between identities and their enrollments will be removed when the option '--reset' is set.»

I would reword it to:

«If the the option '--reset' is set, previous relationships between identities and their enrollments will be removed»

sduenas:
For big databases it will last a little longer but I hope it won't be a problem.

PR #96, opened 2017-10-04
Install misc scripts

sduenas:
During Sorting Hat installation, the scripts stored under 'misc'
folder will be installed in the system.


albertinisg:
Thanks @sduenas!!

PR #100, opened 2017-11-10
[tests] Adapt to testing using MariaDB as backend

jgbarah:
MariaDB has some peculiarities when using it from SQLAlchemy. These changes address them. Most of the changes are for the tests themselves, except for the one to database.py, which includes a new exception which is raised by MariaDB.


tests/test_cmd_organizations.py
```diff
@@ -45,29 +45,37 @@
...
-Example\texample.org
 Example\texample.net
+Example\texample.org
 MyExample\tmyexample.com"""
 
 REGISTRY_OUTPUT_EXAMPLE_ALT = """Example\texample.com
 Example\texample.net"""
 
 
+def sort_lines(lines):
```
sduenas: Instead of doing this, what you have to do is sort the results in `orgs` command. Thus, users will have the same output, no matter which DBMS they are using.

jgbarah: Good idea. I will have a look to learn how to do that.

jgbarah: I've been looking at this. There are two chances:

* sorting in `cmd/organizations.py`. The code is:

```python
    def registry(self, term=None):
       ....
            orgs = api.registry(self.db, term)
            self.display('organizations.tmpl', organizations=orgs)
```

It would be adding `sorting` to orgs:

```python
    def registry(self, term=None):
       ....
            orgs = sorting(api.registry(self.db, term))
            self.display('organizations.tmpl', organizations=orgs)
```

Or instead, go to the real source, which is in `api.py`:

```
def registry(db, term=None):
    ...
        if term:
            orgs = session.query(Organization).\
                filter(Organization.name.like('%' + term + '%')).\
                order_by(Organization.name).all()
           ...
        else:
            orgs = session.query(Organization).\
                order_by(Organization.name).all()
        ...
    return orgs
```

Curiously enough, results (`orgs`) seem to be already ordered. Assuming that for some reason the ordering by MySQL is not the same as for MariaDB, the fix would be letting Python sort instead of the database:

```
def registry(db, term=None):
    ...
        if term:
            orgs = session.query(Organization).\
                filter(Organization.name.like('%' + term + '%')).\
                all()
           ...
        else:
            orgs = session.query(Organization).\
                all()
        ...
    return sorted(orgs)
```

But this seems to break other tests (maybe assuming a certain order in orgs?). I'm not sure if that's the reason or there is something else. Before proceeding, what do you think is more appropriate?

jgbarah: Hmmm. Unfortunately it is not so easy. `orgs` in the code above (in both cases) are not strings, but Organization objects. That means they cannot be sorted. So, either I define a sorting method for them, or we forget about these solutions.

So, now I see three options:

* Leave it as it is, maybe with an accompanying ticket about finding out how to produce the same order with MySQL and MariaDB.
* Explore why both databases give a different order, and fix the code (instead of sorting in the test).
* Define ordering methods for Organization, and use them instead of sorted().

I would vote for the first one, since this was just trying to make the tests run, but I will try whatever you find convenient.

jgbarah: Not completely sure, but maybe this has something to do...: [Why is ORDER BY in a FROM Subquery Ignored?](https://mariadb.com/kb/en/library/why-is-order-by-in-a-from-subquery-ignored/).

sduenas: I've been checking this. I think the problem is not in the way organizations are sorted but in the way their domains are sorted. I think the best way to fix it is to define how the relationship of `Organization` and `Domain` is sorted:  http://docs.sqlalchemy.org/en/latest/orm/extensions/orderinglist.html

jgbarah:
Please, check now. I did the ordering thing for domains, and adapted the tests to the new order (should be the same in MySQL, I checked with MariaDB).

jgbarah:
Weird. Tests pass in my setup (with MariaDB), but Travis triggers some errors in the tests that I cannot reproduce, and at first sight, don't seem to be related to the changes I did... Any idea?

sduenas:
Those errors are related with your change. You have changed the other in the domains, which means those tests that checked domains stuff fail because they aren't sort following the old way.

jgbarah:
@sduenas, but the tests pass in my laptop, with MariaDB. I guess that means that the order in MariaDB is as is in the tests, while in MySQL something happens... That would mean that the fix doesn't work, because it was supposed to work the same way in MariaDB and MySQL. Or am I missing something?

tests/test_utils.py
```diff
@@ -117,6 +117,17 @@ def test_merge_date_ranges(self):
         self.assertEqual(len(ranges), 1)
         self.assertEqual(ranges[0], (datetime.datetime(2000, 1, 1), datetime.datetime(2100, 1, 1)))
 
+        # Case 10
+        dates = [(datetime.datetime(2006, 1, 1), datetime.datetime(2008, 1, 1)),
+                 (datetime.datetime(1999, 1, 1), datetime.datetime(2000, 1, 1)),
+                 (datetime.datetime(1900, 1, 1), datetime.datetime(2100, 1, 1))]
+
+        ranges = [r for r in merge_date_ranges(dates)]
+        print(ranges)
```
sduenas: Please remove this `print`.

jgbarah: Done. Sorry for not removing it earlier.

Thanks!

jgbarah:
Comment addressed and commit rebased.

PR #101, opened 2017-11-10
[setup] Add code for setup.py test to work

jgbarah:


setup.py
```diff
@@ -23,10 +23,13 @@
 
 import codecs
 import os
+import os.path
```
sduenas: This isn't needed. `os` import do the same work.

setup.py
```diff
@@ -90,5 +111,8 @@
         'pandas>=0.17',
         'pyyaml>=3.12'
       ],
+      cmdclass = cmdclass,
+#      tests_require=[],
```
sduenas: Remove this comments.

setup.py
```diff
@@ -49,6 +52,24 @@
...
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        test_suite = unittest.TestLoader().discover('.', pattern='test*.py')
+        result = unittest.TextTestRunner(buffer=True).run(test_suite)
+        sys.exit(not result.wasSuccessful())
+
+cmdclass = {'test':TestCommand}
```
sduenas: This doesn't follow PEP8 specification.

jgbarah:
Oooops. I thought it was self-explanatory. `setup.py test` was not working, now it works...

If you want me to elaborate, that means that now we can run tests as they are usually run when using the setup.py machinery, and they still run when running `run_tests.py` by hand. I'm going to include some of this in the commit, so that it is more clear.

Thanks for the comments, I will address them asap.

setup.py
jgbarah: You're right, sorry.

setup.py
jgbarah: Fixed. I also fixed some other issues reported by flake8.

setup.py
jgbarah: Done.

PR #108, opened 2017-12-05
[sortinghat] Upgrade package dependencies

valeriocos:
This patch sets the minimum versions of PyMySQL and sqlalchemy to 0.7.0 and 1.1.15, respectively.


jgbarah:
Looks good to me. Any comment, @sduenas ?

sduenas:
In any case, this PR doesn't fix the problems with the tables that aren't created with the collation field. The bug persists.

PR #116, opened 2018-01-24
[init] Add option for reusing database

jgbarah:
Really, for raising a new exception when intending to create a new database (using init). And a new option to Init, reuse, that will not trigger printing the message about "error creating database" which appears now every time you call Init.

### DIFF
--- a/sortinghat/exceptions.py
+++ b/sortinghat/exceptions.py
@@ -40,7 +40,7 @@ CODE_LOAD_ERROR = 7
 CODE_MATCHER_NOT_SUPPORTED_ERROR = 8
 CODE_NOT_FOUND_ERROR = 9
 CODE_VALUE_ERROR = 10
-
+CODE_DATABASE_EXISTS = 11
 
 class BaseError(Exception):
     """Base class error.
@@ -90,7 +90,14 @@ class DatabaseError(BaseError):
     """Exception raised when a database error occurs"""
 
     message = "%(error)s (err: %(code)s)"
-    code = CODE_DATABASE_ERROR
+    code = CODE_DATABASE_EXISTS
+
+
+class DatabaseExists(BaseError):
+    """Exception raised when trying to create a database and it already exists"""
+
+    message = "%(error)s (err: %(code)s)"
+    code = CODE_DATABASE_EXISTS
 
 
 class InvalidDateError(BaseError):
--- a/tests/test_cmd_init.py
+++ b/tests/test_cmd_init.py
@@ -36,7 +36,8 @@ from sortinghat import api
 from sortinghat.command import CMD_SUCCESS
 from sortinghat.cmd.init import Init
 from sortinghat.db.database import Database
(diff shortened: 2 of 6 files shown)


tests/test_cmd_init.py
```diff
@@ -114,14 +113,23 @@ def test_existing_db_error(self):
...
+        self.assertEqual(code2, CODE_DATABASE_EXISTS)
 
         # Context added to catch deprecation warnings raised on Python 3
         with warnings.catch_warnings(record=True):
             output = sys.stderr.getvalue().strip()
             self.assertRegexpMatches(output,
                                      DB_EXISTS_ERROR % {'database' : self.name})
 
+    def test_existing_db_reuse(self):
+        """Check if it returns an error when tries to create the registry twice"""
```
sduenas: The description of this test is wrong. Calling to run doesn't return an error in any of the cases.

tests/test_cmd_init.py
```diff
@@ -166,14 +175,31 @@ def test_existing_db_error(self):
...
             self.assertRegexpMatches(output,
                                      DB_EXISTS_ERROR % {'database' : self.name})
 
+        cmd = Init(user=self.kwargs['user'], password=self.kwargs['password'],
+                    host=self.kwargs['host'], port=self.kwargs['port'])
+        code3 = cmd.initialize(self.name)
+        self.assertEqual(code2, CODE_DATABASE_EXISTS)
+
+    def test_existing_db_reuse(self):
+        """Check if it returns an error when tries to create the registry twice"""
```
sduenas: The description of this test is wrong.

tests/test_cmd_init.py
```diff
@@ -166,14 +175,31 @@ def test_existing_db_error(self):
...
-        self.assertEqual(code2, CODE_DATABASE_ERROR)
+        self.assertEqual(code2, CODE_DATABASE_EXISTS)
 
         # Context added to catch deprecation warnings raised on Python 3
         with warnings.catch_warnings(record=True):
             output = sys.stderr.getvalue().strip()
             self.assertRegexpMatches(output,
                                      DB_EXISTS_ERROR % {'database' : self.name})
 
+        cmd = Init(user=self.kwargs['user'], password=self.kwargs['password'],
```
sduenas: Why do you need to create a new `Init` instance? I think assertions from above cover this case. In my opinion, this was already checked.

sortinghat/exceptions.py
```diff
@@ -90,7 +90,14 @@ class DatabaseError(BaseError):
     """Exception raised when a database error occurs"""
 
     message = "%(error)s (err: %(code)s)"
-    code = CODE_DATABASE_ERROR
+    code = CODE_DATABASE_EXISTS
+
+
+class DatabaseExists(BaseError):
```
sduenas: I miss test cases for this exception. Check `test_exceptions.py` file. Test are straightforward.

jgbarah:
All your comments hopefully addressed.

PR #125, opened 2018-02-14
[tests] Create tests database before running tests.

jgbarah:
I got a bit fed up of having to create the database for tests before actually running the tests. And anyway, for automation it is nice that you just "run_tests.py". So, I instrumented the base class for all database tests, so that it creates (only once in a test run) the database before running all the tests.

Since now it is not needed, I also removed the dabase creation line from the travis config.

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -29,12 +29,15 @@ import unittest
 if not '..' in sys.path:
     sys.path.insert(0, '..')
 
-from sortinghat.db.database import Database
+from sortinghat.db.database import Database, DatabaseExists
 
 CONFIG_FILE = 'tests.conf'
 
 
 class TestDatabaseCaseBase(unittest.TestCase):
+
+    db_created = False
+
     """Defines common setup and teardown methods for database-based tests.
 
     When a class inherits from this one and it is instantiated,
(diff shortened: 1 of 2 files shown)


tests/base.py
```diff
@@ -29,12 +29,15 @@
...
 
-from sortinghat.db.database import Database
+from sortinghat.db.database import Database, DatabaseExists
 
 CONFIG_FILE = 'tests.conf'
 
 
 class TestDatabaseCaseBase(unittest.TestCase):
+
+    db_created = False
```
sduenas: This should go after the docstring of this class.

tests/base.py
```diff
@@ -55,6 +62,12 @@ def setUpClass(cls):
                   'database': config['Database']['name'],
                   'host': config['Database']['host'],
                   'port': config['Database']['port']}
+        if not TestDatabaseCaseBase.db_created:
+            try:
+                Database.create(**cls.db_kwargs)
+            except DatabaseExists as e:
+                pass
+            TestDatabaseCaseBase.db_created = True
```
sduenas: Why is this needed? The method `setUpClass` is only called once. Moreover, tests suite must keep the system as it was before running the tests. I think the database should be deleted once the tests finish (dropping the database inside `tearDownClass` method.

PR #133, opened 2018-03-07
[model] Fix wrong __repr__ in Profile

sduenas:
Invalid data were given for the representation of a Profile object.


sduenas:
@acs you're totally right, I was thinking that the gender had its own class. I'll update the PR to keep it as it was but removing the extra parameter that makes this fail.

Issue #144, opened 2018-06-12
Resume method for unify command


sduenas:
Due to `unify` operation is a long running task for big databases, it would be ideal to have checkpoints in case of failure. Therefore, the process would be able to restart from the point it stopped.

Unify consists on two phases. First, matches are calculated in memory. Second, each match is ensured in the database merging identities. It would be possible to store these matches in a temporary file of table, removing entries once they're applied to the database. Thus, if the process fails, it will start again reading and writing those entries instead of start over calculating the matches.

valeriocos:
I have just pushed a PR with the purpose of kickstarting a discussion to address this enhancement. @sduenas, please have a look at it.

sduenas:
As we discussed, we have to get an agreement before continuing working on this. The main problem is how the command should work when multiple databases are in use. In this case, a common file won't work. It will need a specific file for each database. One option would be to store matching data in a temporary table on the database.

If we think in sortinghat working as a server, I think is a better idea to store intermediate results into files. One file for each database. The only problem is where to store the files.

What do you think, @valcos?

valeriocos:
With the current implementation, If SortingHat is executed from a docker container, there will be no problems (as confirmed with the operation team). Every intermediate file will exist in the container and if need, we can persist the files in the file system in a way that they will not clash each other (for instance using the name of the db). This solution is similar to the file logs of other GrimoireLab components (perceval, mordred, etc.).

If we consider SortingHat like a server, I'm not really in favor of polluting the database schema with tables that don't belong to the domain model, nevertheless we can take this path if needed.  I prefer to store intermediate results into files, however this implies that we need a unified format to transparently read the results from the database and files (as done in PR #146 with the method `jsonize`), otherwise we will have duplicated logic (as shown in PR #152). 
The files could be stored in a common directory (e.g., ~/.sortinghat.d/) where their names could be the database names.

Anyway, I'm open to any solution you propose. In case we don't find a solution that makes us happy, we can always postpone this enhancement and move the focus on other tasks.

sduenas:
> If we consider SortingHat like a server, I'm not really in favor of polluting the database schema with tables that don't belong to the domain model, nevertheless we can take this path if needed. I prefer to store intermediate results into files, however this implies that we need a unified format to transparently read the results from the database and files (as done in PR #146 with the method jsonize), otherwise we will have duplicated logic (as shown in PR #152).
[...]

PR #146, opened 2018-06-14
[unify] Add resume method to unify command

valeriocos:
This code addresses #144. Thus, it proposes a method to recover failed unify executions. In a nutshell, after calculating the matches in memory, they are converted to JSON items and marked as not processed. Each match successfully executed is marked as processed. In case of a failure, the failed match plus the remaining ones are serialized to disk (which location can be modified by command line). In the next execution, those matches will be loaded and re-executed again.


sortinghat/cmd/unify.py
```diff
@@ -23,13 +23,17 @@
...
+import json
+import os
 
 from .. import api
 from ..command import Command, CMD_SUCCESS, HELP_LIST
 from ..exceptions import MatcherNotSupportedError
 from ..matcher import create_identity_matcher, match
 from ..matching import SORTINGHAT_IDENTITIES_MATCHERS
 
+MATCHES_PATH = '~/.sortinghat/matches/matches.log'
```
sduenas: I don't think we need a subdir for this file. Actually, I'd rather name the archive to the command which is running it. In this case, `unify`.

sortinghat/cmd/unify.py
```diff
@@ -58,6 +62,8 @@ def __init__(self, **kwargs):
                                  help="do not rigorous check of values (i.e, well formed email addresses)")
         self.parser.add_argument('-i', '--interactive', action='store_true',
                                  help="run interactive mode while unifying")
+        self.parser.add_argument('-l', '--log-path', dest='log_path', default=None,
```
sduenas: In my opinion, the user shouldn't be able to define the matches file. I think is something internal to this command. What the user can do is whether it want to use the resume/recovery mode or not.
The file should be always stored in the same place.

sortinghat/cmd/unify.py
```diff
@@ -85,13 +92,13 @@ def run(self, *args):
...
                           params.fast_matching, params.no_strict,
-                          params.interactive)
+                          params.interactive, params.log_path)
 
         return code
 
     def unify(self, matching=None, sources=None,
               fast_matching=False, no_strict_matching=False,
-              interactive=False):
+              interactive=False, log_path=None):
```
sduenas: `log_path` is not a clear name for this parameter. For its name, looks like it's doing something different.

sortinghat/cmd/unify.py
```diff
@@ -120,9 +127,18 @@ def unify(self, matching=None, sources=None,
         :param fast_matching: use the fast mode
         :param no_strict_matching: disable strict matching (i.e, well-formed email addresses)
         :param interactive: interactive mode for merging identities
+        :param log_path: path of the log containing matches
         """
         matcher = None
 
+        if not log_path:
```
sduenas: When I thought about adding this feature I thought of different behaviour:

* When a recovery or resume file is not given, the command works as always
* When a file is given:
    * If it doesn't exist, after the matching process, identities will be written to that file
    * If it exists, it means something went wrong the last time this process run, so it will skip matching process and will start merging identities.

What do you think? Don't you think is better this way?

sortinghat/cmd/unify.py
```diff
@@ -216,3 +243,66 @@ def __display_stats(self):
         self.display('unify.tmpl', processed=self.total,
                      matched=self.matched,
                      unified=self.total - self.matched)
+
+    def __matches_log(self):
```
sduenas: All these methods can be encapsulated inside an object. This object will save, remove or recover entries and it is independent of how the matches are stored under the hood.

tests/test_cmd_unify.py
```diff
@@ -188,6 +251,126 @@ def test_unify(self):
...
+
+        with open(self.custom_log, 'r') as f:
+            count_objs = 0
+            for line in f.readlines():
+                matches_obj = json.loads(line.strip("\n"))
+                self.assertTrue(all([isinstance(m, six.string_types) for m in matches_obj['identities']]))
+                self.assertFalse(matches_obj['processed'])
+                count_objs += 1
+
+                self.assertTrue(count_objs, 5)
```
sduenas: There's a mistake here. This assertion should be outside of the loop. Actually, it's hiding an error. The method to use is `assertEqual` and not `assertTrue`. That's why I don't like to use `assertTrue`.

tests/test_cmd_unify.py
```diff
@@ -188,6 +251,126 @@ def test_unify(self):
...
+
+        with open(self.custom_log, 'r') as f:
+            count_objs = 0
+            for line in f.readlines():
+                matches_obj = json.loads(line.strip("\n"))
+                self.assertTrue(all([isinstance(m, six.string_types) for m in matches_obj['identities']]))
+                self.assertFalse(matches_obj['processed'])
+                count_objs += 1
+
+        self.assertTrue(count_objs, 5)
```
sduenas: Same error as above here.

sortinghat/cmd/unify.py
valeriocos: ok, I'll give it a try

valeriocos:
Sure, I have just updated the PR with your suggestions

sortinghat/cmd/unify.py
```diff
@@ -117,13 +124,16 @@ def unify(self, matching=None, sources=None,
         :param fast_matching: use the fast mode
         :param no_strict_matching: disable strict matching (i.e, well-formed email addresses)
         :param interactive: interactive mode for merging identities
+        :param recovery: if enabled, the unify will read the matching identities stored in
+           recovery file (`~/.sortinghat/unify.log`) and process them
```
sduenas: This is not the file where data is stored. I think it is better to say `stored in recovery file (RECOVERY_FILE_PATH)`

sortinghat/cmd/unify.py
```diff
@@ -155,22 +165,40 @@ def __unify_unique_identities(self, uidentities, matcher,
         self.total = len(uidentities)
         self.matched = 0
 
-        matched = match(uidentities, matcher, fastmode=fast_matching)
+        if self.recovery and RecoveryFile.exists():
+            print("Loading matches from recovery file: %s" % RecoveryFile.location())
+            matched = RecoveryFile.load_matches()
+        else:
```
sduenas: This `if` block means `recovery` is always in use. This code should be under an `if` too.

```
 matched = RecoveryFile.jsonize(matched)
```

valeriocos: The `jsonize`method converts the matched identities to a common JSON format, thus it simplifies the logic to deal with matches collected from the recovery file or with the method `match`. However that method will be always used (even if the recovery is not active). 
An option could be to move that method from the `RecoveryFile`class to the `Unify` class. Otherwise, I guess the PR should be significantly changed (see PR #152)

sduenas:
Merged

PR #147, opened 2018-06-18
Low-level database API

sduenas:
This PR provides a low-level API to execute basic operations over the database, such as adding or removing identities or finding entities. All these operations work with a session. Nothing is stored in the database until the session is closed. Thus, these functions can be considered as "bricks", that combined can create high-level functions.


sortinghat/db/api.py
```diff
@@ -0,0 +1,656 @@
...
+    session.add(enrollment)
+
+    return enrollment
+
+
+def withdraw(session, uidentity, organization,
+             from_date=MIN_PERIOD_DATE, to_date=MAX_PERIOD_DATE):
+    """Withdraw a unique identity from an organization.
+
+    Removes all the enrollments between the unique identity in
```
valeriocos: something missing after the word between?

.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
   - pip install python-dateutil>=2.6.0
   - pip install pandas==0.18.1
   - pip install requests>=2.9
-  - pip install httpretty==0.8.6
+  - pip install httpretty
```
jgbarah: Interesting. This means that we no longer need to specify `httpretty==0.8.6`? If that's the case, shouldn't we specify `>=something`?

sortinghat/cmd/load.py
```diff
@@ -423,11 +423,16 @@ def __load_identities(self, identities, uuid, verbose):
     def __load_profile(self, profile, uuid, verbose):
         """Create a new profile when the unique identity does not have any."""
 
+        def is_empty_profile(prf):
```
valeriocos: why is_empty_profile is inside __load_profile?

sortinghat/db/api.py
```diff
@@ -0,0 +1,656 @@
...
+    session.add(organization)
+
+    return organization
+
+
+def delete_organization(session, organization):
+    """Remove an organization from the session.
+
+    Function that removes from the session the organization
+    given in `organization`. Data related  such as domains
```
valeriocos: too many blanks after related?

.travis.yml
sduenas: Yes, you're probably right. We can use the latest version of `httpretty` (0.9.5).

sortinghat/cmd/load.py
sduenas: It's a private function for that method. Only used there and I don't think other methods will need it.

.travis.yml
jgbarah: In that case, I suggest that you change that line to

`pip install httpretty>=0.9.5`

PR #152, opened 2018-06-22
[unify] Add resume method to unify command (2)

valeriocos:
This PR proposes a different implementation of the PR #146


valeriocos:
After discussion with @sduenas , I close this PR and keep working on #146

Issue #149, opened 2018-06-19
Add logging system


sduenas:
Sorting Hat lacks of a logging system. We can implement the same system we already have in other GrimoireLab tools such as Perceval.

PR #153, opened 2018-06-22
[sortinghat] Add logging support

valeriocos:
This PR targets issue #149 . Thus it includes support to for logging.

### DIFF
--- a/sortinghat/command.py
+++ b/sortinghat/command.py
@@ -19,12 +19,15 @@
 #     Santiago Dueñas <sduenas@bitergia.com>
 #
 
+import logging
 import sys
 
 import jinja2
 
 from .exceptions import DatabaseError
 from .db.database import Database
+
+logger = logging.getLogger(__name__)
 
 CMD_SUCCESS = 0
 CMD_FAILURE = 1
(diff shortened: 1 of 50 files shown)


sortinghat/command.py
```diff
@@ -57,20 +60,22 @@ def display(self, template, **kwargs):
...
         s = t.render(**kwargs)
-
+        logger.info(s)
+        # this is left to not break the tests
         sys.stdout.write(s)
 
     def error(self, msg):
         s = "Error: %s\n" % msg
-        sys.stderr.write(s)
+        logger.error(s)
```
sduenas: I wouldn't change this. We have to make a difference between those messages related to the internal work of the tool (log messages) and those messages generated as an output for the user (which is this case).

valeriocos: ok @sduenas, I'm going to revert this change and update the PR

valeriocos:
I have updated the pr according to your comments @sduenas , please have a look

sortinghat/command.py
```diff
@@ -57,7 +60,8 @@ def display(self, template, **kwargs):
 
         t = env.get_template(template)
         s = t.render(**kwargs)
-
+        logger.info(s)
```
sduenas: This should be also removed as you did with the errors.

valeriocos:
I have updated the PR, including changes due to the inclusion of #155 to the codebase.

PR #156, opened 2018-07-26
[db] Handle disconnection using pessimistic mode

sduenas:
SQLAlchemy offers a pessimistic mode to handle database disconnection. Setting 'pool_pre_ping' parameter on the database engine will check if the database connection is still active when a session of the connection pool is reused. To check that it runs a simple SQL statement. This causes a small hit in the performance but it's worth it.

This feature was included in 1.2 version, so the list of dependencies has been updated as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ $ sortinghat load identities.json
 
 * Python >= 3.4
 * MySQL >= 5.5
-* SQLAlchemy >= 1.0.0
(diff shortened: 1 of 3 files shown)


dpose:
@sduenas 
I've tested it and it is working successfully. No more 'mysql timeouts' nor 'mysql server has gone away' after about a day.
LGTM.

PR #157, opened 2018-08-04
Print uuid of import errors to allow investigating

aklapper:


sortinghat/utils.py
```diff
@@ -150,11 +150,11 @@ def uuid(source, email=None, name=None, username=None):
         parameters is None; parameters are empty.
     """
     if source is None:
-        raise ValueError("source cannot be None")
+        raise ValueError("source cannot be None for source %s" % source)
```
sduenas: For those cases when `source` is `None` or an empty string, does it make sense to print it?

aklapper: Hmm. You're probably right. Maybe that was the one case that I did not actually run into. :)

sduenas:
This PR is incompatible. The current version of SortingHat fixes this problem.

PR #168, opened 2018-11-28
[cmd:autoprofile] Avoid to set email addresses in name profile

sduenas:
When there are email address in 'name' or 'username' identity fields is possible that name profile will be set to this value.This can cause privacy issues in some contexts. For this reason, autoprofile will not set email values in this field.


valeriocos:
LGTM

PR #169, opened 2018-12-11
[model] Migrate model to Django ORM

### LINKED ISSUE #166: Replace SQLAlchemy by Django ORM
Django is the framework selected to turn Sorting Hat into a web service. For this reason, it makes sense to get rid off SQLAlchemy and use the ORM defined by Django.

sduenas:
The previous Sorting Hat model was based on SQLAlchemy ORM. This one preserves the same structure but adds two new fields to each object. These two new fields are 'created_at' and 'last_modified'.

The first field will be set only to the date when an object is created. The second one will be set automatically when an object is modified.

Along with this commit, it's been included the migrations file used by Django to setup the database; the requirements file with the Python packages needed to run this software and a directory with unit tests.

To run the tests you'll need to execute the next command from the root directory:

```
./manage.py test --settings=config.settings.testing
```

This PR fixes #166 .


sduenas:
@valeriocos please review this PR. Once it is reviewed and tested I'll merge it.

Issue #178, opened 2019-01-08
Implement GraphQL mutations to add identities


sduenas:
The goal of this task is to implement basic mutations to add identities. This implies to create a default profile.

When a new identity is added, either it would create a unique identity (if it doesn't exist) or it would be linked to an existing unique identity.

sduenas:
Done. Closed by #180.

PR #180, opened 2019-01-16
Define mutation to add identities

sduenas:
The mutation `addIdentity` allows to add a new identity to the registry. These identities can be added to
an existing unique identity. By default, a new unique identity will be created which each new identity.

Along with this mutation, a set of functions have been ported from the previous version of SortingHat, like the generator of **UUIDs**, useful exceptions or the API handlers to add identities in an atomic transaction.

This PR addresses the requirements of #178 .

### DIFF
--- a/config/settings/testing.py
+++ b/config/settings/testing.py
@@ -15,12 +15,25 @@ INSTALLED_APPS = [
     'sortinghat.core',
 ]
 
+SQL_MODE = [
+    'ONLY_FULL_GROUP_BY',
+    'NO_ZERO_IN_DATE',
+    'NO_ZERO_DATE',
+    'ERROR_FOR_DIVISION_BY_ZERO',
+    'NO_AUTO_CREATE_USER',
+    'NO_ENGINE_SUBSTITUTION',
+]
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.mysql',
         'USER': 'root',
         'PASSWORD': '',
         'NAME': 'sortinghat_db',
+        'OPTIONS': {
+            'charset': 'utf8mb4',
+            'sql_mode': ','.join(SQL_MODE)
+        },
         'TEST': {
             'NAME': 'testhat',
             'CHARSET': 'utf8mb4',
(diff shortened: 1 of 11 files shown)


sortinghat/core/db.py
```diff
@@ -25,8 +25,34 @@
...
+    :returns: a unique identity object
+
+    :raises NotFoundError: when the unique identity with
+        the given `uuid` does not exists.
+    """
+    try:
+        uidentity = UniqueIdentity.objects.get(uuid=uuid)
+    except UniqueIdentity.DoesNotExist:
+        raise NotFoundError(entity=uuid)
+    else:
```
valeriocos: is this `else` needed?

config/settings/testing.py
```diff
@@ -21,6 +21,9 @@
         'USER': 'root',
         'PASSWORD': '',
         'NAME': 'sortinghat_db',
+        'OPTIONS': {
+            'charset': 'utf8mb4'
```
valeriocos: when testing the PR, I'm getting the following error: 
```
ERROR: test_duplicate_identities_with_truncated_values (tests.test_api.TestAddIdentity)
Check if the same identity with truncated values is not inserted twice
...
mysql_exceptions.DataError: (1406, "Data too long for column 'email' at row 1")
```

To fix it, I would add to `OPTIONS` the following entry to to swtich MySQL mode to not STRICT
`'sql_mode': 'NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION'`. What do you think?

sortinghat/core/db.py
sduenas: Not really but I included it because I didn't want `uidentity` was used outside the `try...except` block.

config/settings/testing.py
sduenas: I got that error but I disabled it doing something similar in my MySQL config. This makes more sense. I will add some other modes that I think they should still there.

sduenas: I've just uploaded some changes that fix it.

sduenas:
@valeriocos done, please review my comments

Issue #179, opened 2019-01-08
Implement GraphQL mutations to delete identities


sduenas:
The goal of this task is to implement basic mutations to remove identities. Take into account removing an identity would mean to remove a unique identity too.

sduenas:
Fixed by #183

PR #183, opened 2019-01-21
Define mutation to remove identities

sduenas:
The mutation `deletedentity` allows to remove identities and unique identities from the registry.

This PR addresses the requirements of #179 .

### DIFF
--- a/sortinghat/core/db.py
+++ b/sortinghat/core/db.py
@@ -257,6 +257,19 @@ def add_identity(uidentity, identity_id, source,
     return identity
 
 
+def delete_identity(identity):
+    """Remove an identity from the database.
+
+    This function removes from the database the identity given
+    in `identity`. Take into account this function does not
+    remove unique identities in the case they get empty.
+
+    :param identity: identity to remove
+    """
+    identity.delete()
+    identity.uidentity.save()
+
+
 _MYSQL_DUPLICATE_ENTRY_ERROR_REGEX = re.compile(r"Duplicate entry '(?P<value>.+)' for key")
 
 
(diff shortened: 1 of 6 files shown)


sortinghat/core/db.py
```diff
@@ -55,6 +55,28 @@ def find_unique_identity(uuid):
         return uidentity
 
 
+def find_identity(uuid):
+    """Find an identity.
+
+    Find an identity by its UUID in the database. When the
+    identity does not exist the function will raise
+    a `NotFoundException`.
```
valeriocos: `NotFoundException` should be `NotFoundError`

tests/test_api.py
```diff
@@ -467,3 +471,132 @@ def test_none_or_empty_data(self):
...
+
+        jrae = api.add_identity('scm',
+                                name='Jane Rae',
+                                email='jrae@example')
+        Enrollment.objects.create(uidentity=jrae.uidentity,
+                                  organization=libresoft_org)
+
+    def test_delete_identity(self):
+        """Check whether it deletes an identity"""
+
```
valeriocos: I would add some code to show the number of initiali unique identities and identities. SOmething like:
```
uidentities = UniqueIdentity.objects.all()
 self.assertEqual(len(uidentities), 4)

identities = Identity.objects.all()
self.assertEqual(len(identities), 4)
```

tests/test_api.py
```diff
@@ -467,3 +471,132 @@ def test_none_or_empty_data(self):
...
+        identities = Identity.objects.all()
+        self.assertEqual(len(identities), 3)
+
+        with self.assertRaises(ObjectDoesNotExist):
+            Identity.objects.get(id='1387b129ab751a3657312c09759caa41dfd8d07d')
+
+    def test_delete_unique_identity(self):
+        """Check whether it deletes a unique identity when its identifier is given"""
+
```
valeriocos: same comment as above

Issue #184, opened 2019-01-21
GraphQL mutation to edit/update the profile of a unique identity


sduenas:
The goal is to be able to edit unique identity profiles to set the display name or email of an identity, the gender, the country of origin or whether an identity is a bot or not.

sduenas:
Fixed by #189

PR #189, opened 2019-01-24
Add GraphQL mutation to edit/update profiles

sduenas:
This mutation allows to update profile identities. Some functions were added to the API to update profiles in the database layer.

This commit adds the features requested on #184

### DIFF
--- a/sortinghat/core/api.py
+++ b/sortinghat/core/api.py
@@ -28,7 +28,8 @@ from .db import (find_unique_identity,
                  add_unique_identity as add_unique_identity_db,
                  add_identity as add_identity_db,
                  delete_unique_identity as delete_unique_identity_db,
-                 delete_identity as delete_identity_db)
+                 delete_identity as delete_identity_db,
+                 update_profile as update_profile_db)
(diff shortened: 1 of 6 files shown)


sortinghat/core/api.py
```diff
@@ -201,3 +202,47 @@ def delete_identity(uuid):
...
+    This function allows to edit or update the profile information
+    of the unique identity identified by `uuid`.
+
+    The values to update are given as keyword arguments. The allowed
+    keys are listed below (other keywords will be ignored):
+
+       - 'name' : name of the unique identity
+       - 'email' : email address of the unique identity
+       - 'gender' : gender of the unique identity
+       - 'gender_acc' : gender accuracy (range of 0 to 100; by default,
```
valeriocos: According to the tests, this should be:
`gender accuracy (range of 1 to 100; by default, set to 100)`

sduenas: Fixed!

valeriocos: thanks!

valeriocos:
LGTM

PR #190, opened 2019-01-25
Allow to search uidentities by an uuid

dlumbrer:
In order to get one uidentity by its uuid we added a filter to the query. We had to install a graphene-django-extras[1] library to do that in the simplest way. There are three behaviors:

1. If you specify the uuid in the filter you will receive the uidentites that match that uuid.
2. If you don't put the uuid filter, you will receive all the uidentites.
3. If you put an uuid in the filter that does not exist, you will receive an empty list.

There are also tests in order to check that it works as expected.

[1]: https://github.com/eamigo86/graphene-django-extras

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,8 @@
 Django==2.1
+django-filter>=2
 grimoirelab_toolkit>=0.1.8
 mysqlclient==1.3.13
 python-dateutil>=2.6.0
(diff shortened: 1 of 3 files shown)


dlumbrer:
Finally, we decided to not use the graphene-django-extras api, we followed another way to do the filter (with Django). Closing this PR for #193

PR #196, opened 2019-02-28
[tests] Check whether last_modified field is updated in withdraw() call

sduenas:
This test was missing in the PR that added this functionality. It tests whether the last date of modification for a unique identity is updated when withdraw() function is called.


sduenas:
@valeriocos hope you can review this PR

Issue #162, opened 2018-10-16
Do not allow to insert values composed only by white spaces


sduenas:
SortingHat API (not the command line client) allows to insert values only formed by white spaces in the database. These values aren't valid so it's usage should be forbidden.

PR #210, opened 2019-09-04
[utils] Prevent insertion of white space values

valeriocos:
This PR addresses https://github.com/chaoss/grimoirelab-sortinghat/issues/162, thus it processes the values of email, name and username by splitting white spaces.

Tests have been addded accordingly.

### DIFF
--- a/sortinghat/utils.py
+++ b/sortinghat/utils.py
@@ -149,17 +149,21 @@ def uuid(source, email=None, name=None, username=None):
     :raises ValueError: when source is None or empty; each one of the
         parameters is None; parameters are empty.
     """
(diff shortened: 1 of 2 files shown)


valeriocos:
The travis error is not related to this PR

valeriocos:
Related PR: https://github.com/chaoss/grimoirelab-sortinghat/pull/211

sduenas:
I opened #162 to solve the problem in `muggle` branch. Sorry if I didn't make that clear. I think we should leave this version as it is.
In any case, the solution shouldn't be this way. It should be handled before the `uuid` is created. This function shouldn't take care of that.

PR #215, opened 2019-09-13
[db][utils] Prevent insertion of white space values

mafesan:
The method `validate_field` replaces the checks for string values when they are `None` or empty and extends those checks to strings composed only by white spaces (tabs included).

This PR aims to solve issue https://github.com/chaoss/grimoirelab-sortinghat/issues/162.


sortinghat/core/utils.py
```diff
@@ -119,3 +120,32 @@ def unaccent_string(unistr):
...
+    not composed only by whitespaces and/or tabs.
+
+    :param name: Name of the field to validate
+    :param value: Value of the field to validate
+    :param allow_none: `True` if `None` values are permitted, `False` otherwise
+
+    :raises ValueError: when a condition to validate the string is not satisfied
+    :raises TypeError: when the input value is not a string and not `None`
+    """
+    is_string = isinstance(value, str)
```
sduenas: Probably the first thing is to compare with `None` cases, so you don't have to add extra checks all over the code.

sortinghat/core/utils.py
```diff
@@ -119,3 +120,32 @@ def unaccent_string(unistr):
...
+
+
+def validate_field(name, value, allow_none=False):
+    """Validate a given string field following a set of rules
+
+    The conditions to validate `value` consists on checking if its value is `None`
+    and if this is allowed or not; if its value is not an empty string or if it is
+    not composed only by whitespaces and/or tabs.
+
+    :param name: Name of the field to validate
```
sduenas: We start these descriptions with lowercase letters.

sortinghat/core/utils.py
```diff
@@ -119,3 +120,32 @@ def unaccent_string(unistr):
...
+    :param value: Value of the field to validate
+    :param allow_none: `True` if `None` values are permitted, `False` otherwise
+
+    :raises ValueError: when a condition to validate the string is not satisfied
+    :raises TypeError: when the input value is not a string and not `None`
+    """
+    is_string = isinstance(value, str)
+    if not is_string and value is not None:
+        msg = "field value must be a string; {} given".format(value.__class__.__name__)
```
sduenas: We should stick to one way of working with formatting strings. Let's keep the `format` one. We'll add `f-strings` later.

tests/test_db.py
```diff
@@ -740,6 +768,14 @@ def test_identity_id_empty(self):
         with self.assertRaisesRegex(ValueError, IDENTITY_ID_EMPTY_ERROR):
             db.add_identity(uidentity, '', 'scm')
 
+    def test_identity_id_whitespaces(self):
+        """Check whether an identity with an ID composed by whitespaces cannot be added"""
+
+        uidentity = UniqueIdentity.objects.create(uuid=' ')
+
+        with self.assertRaisesRegex(ValueError, IDENTITY_ID_WHITESPACES_ERROR):
+            db.add_identity(uidentity, '\t  ', 'scm')
```
sduenas: This test is different to the other ones. You are comparing `\t    ' when others have only whitespaces. Why don't you add both checks to each test?

tests/test_utils.py
```diff
@@ -395,3 +399,52 @@ def test_no_string(self):
...
             unaccent_string(1234)
+
+
+class TestValidateField(TestCase):
+    """Unit tests for validate_field"""
+
+    def test_validate_string(self):
+        """Check valid and invalid fields"""
+
+        result = validate_field('test_field', 'Test')
```
sduenas: This function returns nothing. I'm not sure if we should compare it with `None`.

sortinghat/core/utils.py
mafesan: Done.

sortinghat/core/utils.py
mafesan: Done, sorry about that.

sortinghat/core/utils.py
mafesan: I agree. I have changed it for using `format` method only.

tests/test_db.py
mafesan: You are right. Now three cases are tested: one with only whitespaces `'  '`, other with tab characters `'\t'` and other case mixing the other two: `'  \t  '`.

tests/test_utils.py
mafesan: Now this method is called twice with valid fields and using no assertions. If the values are valid, no exception has to be raised.

mafesan:
Thank you for your comments @sduenas, I have just addressed them. Please, review the latest changes.

sortinghat/core/utils.py
```diff
@@ -119,3 +120,35 @@ def unaccent_string(unistr):
     string = ''.join(cs)
 
     return string
+
+
+def validate_field(name, value, allow_none=False):
+    """Validate a given string field following a set of rules
```
sduenas:
```suggestion
    """Validate a given string field following a set of rules.
```

tests/test_utils.py
```diff
@@ -395,3 +399,55 @@ def test_no_string(self):
...
+            validate_field('test_field', '\t')
+
+        expected = FIELD_WHITESPACES_ERROR.format(name='test_field')
+        with self.assertRaisesRegex(ValueError, expected):
+            validate_field('test_field', '  \t ')
+
+    def test_allow_none(self):
+        """Check valid and invalid fields allowing `None` values"""
+
+        result = validate_field('test_field', None, allow_none=True)
```
sduenas: You should remove this result too.

tests/test_utils.py
```diff
@@ -395,3 +399,55 @@ def test_no_string(self):
...
+
+    def test_validate_string(self):
+        """Check valid and invalid fields"""
+
+        # If the field is valid, the method does not raise any exception
+        validate_field('test_field', 'Test')
+
+        validate_field('test_field', None, allow_none=True)
+
+        expected = FIELD_EMPTY_ERROR.format(name='test_field')
```
sduenas: Maybe these tests can be split in a different test.

tests/test_utils.py
```diff
@@ -395,3 +399,55 @@ def test_no_string(self):
...
+class TestValidateField(TestCase):
+    """Unit tests for validate_field"""
+
+    def test_validate_string(self):
+        """Check valid and invalid fields"""
+
+        # If the field is valid, the method does not raise any exception
+        validate_field('test_field', 'Test')
+
+        validate_field('test_field', None, allow_none=True)
```
sduenas: This one is not needed because you are testing this parameter in a dedicated test.

tests/test_db.py
```diff
@@ -756,23 +816,73 @@ def test_source_empty(self):
...
+
+        with self.assertRaisesRegex(ValueError, SOURCE_WHITESPACES_ERROR):
+            db.add_identity(uidentity, '1234567890ABCDFE', '\t')
+
+        with self.assertRaisesRegex(ValueError, SOURCE_WHITESPACES_ERROR):
+            db.add_identity(uidentity, '1234567890ABCDFE', ' \t ')
+
     def test_data_none_or_empty(self):
-        """Check whether new identities cannot be added when identity data is None or empty"""
+        """Check whether new identities cannot be added when identity data is None, empty or composed by whitespaces"""
```
sduenas: This comment is too long. Please split it in two lines.

sortinghat/core/utils.py
mafesan: Done.

tests/test_utils.py
mafesan: Oops! My bad. Done.

tests/test_utils.py
mafesan: I agree. I have added a separated test for invalid strings.

tests/test_utils.py
mafesan: Ok.

tests/test_db.py
mafesan: Ok, done. I have changed it following this format:
```python
def my_test(self):
"""
<line1>
<line2>
"""
code
```

Is this what you expect? Thanks in advance!

mafesan:
@sduenas I have addressed you comments. Thanks!

PR #218, opened 2019-10-08
Add history of changes

mafesan:
# Summary 

In this approach I have implemented the following idea about creating a **transaction** inside a given method in the API which will be shared across the individual **operations** performed into the DB.

These are the fields for a given **transaction**:

* `tuid`: Unique identifier for the transaction.
* `timestamp`: Non-editable datetime when the transaction is performed.
* `name`: Name of the method creating the transaction (in this case, it matches with the names from the API, like `add_idenity`, `update_profile`, etc.

The second type of object is **operations**. These are its fields:

* `ouid`: unique identifier for the operation.
* `type`: type of interaction with the DB (`add`, `delete` or `update`).
* `entity`: type of object involved in the operation (`identity`, `organization`, `domain`, etc.).
* `target`: main value of the entity targeted in the operation.
* `timestamp`: datetime when operation was created.
* `args`: serialized input arguments (JSON) from the method performing the operation.
* `trx`: `Transaction` object as a general context where this operation is performed.

About which operations can be applied to which entities, I have prepared the following table:

|                            | Add | Delete | Update |
| ------------- |:-------------:| -----:| -----:|
| Unique identity   |  x     | x          |            |
| Identity               |  x     | x          |  x        |
| Organization       |  x     | x          |            |
| Domain              |  x     | x          |            |
| Enrollment          |  x     | x          |  x        |
| Profile                 |        |             |   x       |
| Blacklist entry\*     |  x     | x          |            |

\* There is still no implementation yet for managing blacklist entries.

The creation of both `Transaction` and `Operation` objects is managed by a new class
named `TransactionsLog`. To create a `TransactionsLog` object it is necessary to call to 
a class method called `open`, which creates a new `Transaction` object saving it into the database.
This will be a unique transaction which will be linked to every `TransactionsLog` object.

Now, to create a new `Operation` object associated with a given transaction, we will call the method
`log_operation` from the `TransactionsLog` object containing the transaction.

Once the transaction is over, it will be closed calling to the `close` method. This means
no more operations can be logged to this transaction.

## Flow example

This feature has been integrated with the rest of the components. This is an example of the flow
when a method in the API is called.

* Starting from a given method in `api.py`, a `TransactionsLog` object is created and passed
as argument when a DB method modifying the database is called. When there are no more DB methods
to call, the transaction can be closed. 

Note: closing a `TransactionsLog` object means adding a `closed_at` datetime
and set the `is_closed` flag to `True` in its inner `Transaction` object.

```python
def my_api_method(arg1, arg2):
    (...)
    trxl = TransactionsLog.open('my_api_method')

    db_method_add(trxl, arg1, arg2)

    trxl.close()
```

* Now, every time a method from `db.py` (related with DB operations modifying the DB in any way, that is any add, delete or update) is called, now those methods
have the `TransactionsLog` object as an additional, mandatory argument.
Inside any of these `db.py` methods performing add/delete/update operations, the steps are:
  * Saving the input arguments before they are modified.
  * After any save/delete operation, logging it creating a **operation** object, passing as arguments the operation type ('add', ' delete', ' update'), the entity
  (`identity', 'domain', etc.), the main value of the targeted entity, a timestamp, the input args and the related transaction.

```python
def db_method_add(trxl, arg1, arg2):

    op_args = {
        "arg1": arg1
        "arg2": arg2
    }

    db_object = Example(arg1=arg1, arg2=arg2)
    db_object.save()

    trxl.log_operation(op_type=Operation.OpType.ADD, entity_type='example',
                       timestamp=datetime_utcnow(), args=op_args,
                       target=op_args['arg1'])
        
```

## Other additions

* (errors) `ClosedTransactionError`: This new error raises when attempting to log an operation 
on a closed transaction.

* (models) `ModelBase` class was renamed to `EntityBase`: This class creates two fields, `created_at` and `last_modified`.
This means every new model inheriting from it will have these two fields by default. So far, these fields have been neeed in all
the models but we could need new ones without these default fields.

* (tests) Tests for the api, db, schema and models were modified to support these changes. There are new tests
for the new `TransactionsLog` class, the new `ClosedTransactionError` and the new `Transaction` and `Operation` models.

* **New dependency**: 
```
django-mysql >= 3.2.0
``` 

# Troubleshooting

These are a few commands which might be useful when trying these changes:

## Execute migration to apply the new models

```shell
$ python3 manage.py makemigrations --settings=config.settings.devel
$ python3 manage.py migrate --settings=config.settings.devel

$ python3 manage.py makemigrations --settings=config.settings.testing
$ python3 manage.py migrate --settings=config.settings.testing
```

## Applying migrations after modifying `models.py`

From the initial set of files in `/sortinghat/core/migrations` before performing any modification, run the command with the `--dry-run` flag to preview the changes to be applied. Example:

```shell
$ python3 manage.py makemigrations --settings=config.settings.devel -n log core --dry-run
```

## Reset Database

```shell
$ python3 manage.py flush --settings=config.settings.devel
```


sortinghat/core/models.py
```diff
@@ -183,3 +185,66 @@ class MatchingBlacklist(ModelBase):
...
+    PROFILE = 'profile'
+    BLACKLIST = 'blacklist'
+
+    OP_TYPE_CHOICES = ((ADD, 'add'), (DELETE, 'delete'), (UPDATE, 'update'))
+    ENTITY_CHOICES = ((UUID, 'unique_identity'), (UID, 'identity'), (ORG, 'organization'),
+                      (DOMAIN, 'domain'), (ENROLLMENT, 'enrollment'), (PROFILE, 'profile'),
+                      (BLACKLIST, 'blacklist_entry'))
+
+    ouid = CharField(max_length=MAX_SIZE_CHAR_FIELD, primary_key=True)
+    type = CharField(max_length=MAX_SIZE_CHAR_FIELD, choices=OP_TYPE_CHOICES, null=False)
```
sduenas: Please check `null` fields because I think by default `null=False` is the defauld behaviour.

mafesan: You are right, I confirm `null=False` is the default behaviour: https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.Field.null .

tests/test_model.py
```diff
@@ -598,3 +601,81 @@ def test_last_modified(self):
...
+
+    def test_unique_transactions(self):
+        """Check whether transactions are unique"""
+
+        with self.assertRaisesRegex(IntegrityError, DUPLICATE_CHECK_ERROR):
+            timestamp = datetime_utcnow()
+            Transaction.objects.create(tuid='12345abcd',
+                                       name=Transaction.ADD_ID, timestamp=timestamp)
+            Transaction.objects.create(tuid='12345abcd',
+                                       name=Transaction.ADD_ID, timestamp=timestamp)
```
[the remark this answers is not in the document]
mafesan: After a later discussion, we agreed on changing the type of `timestamp` field from `CreationDateTimeField` to a simple `DateTimeField` so its value is not initialized by default but with a value passed explicitly.

This timestamp should be created by the method opening the transaction. The same behavior is expected with `Operation` class.

The rationale for this change is that this way we can decide the date-time we are using as a timestamp. For instance, we cannot have the precise timestamp from when a given operation is performed so we set a timestamp immediately after the operation is performed, and eventually could be set to a exact date-time if we could get it somehow.

tests/test_model.py
```diff
@@ -598,3 +601,81 @@ def test_last_modified(self):
...
+
+    def test_created_at(self):
+        """Check creation date is only set when the object is created"""
+
+        before_dt = datetime_utcnow()
+        trx = Transaction.objects.create(tuid='12345abcd',
+                                         name=Transaction.ADD_ID, timestamp=datetime_utcnow())
```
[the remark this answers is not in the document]
mafesan: See my comment above: https://github.com/chaoss/grimoirelab-sortinghat/pull/218#discussion_r335912753.

tests/test_model.py
```diff
@@ -598,3 +601,81 @@ def test_last_modified(self):
...
+
+
+class TestOperation(TransactionTestCase):
+    """Unit tests for Operation class"""
+
+    def setUp(self):
+        """Load initial dataset"""
+
+        Transaction.objects.create(tuid='0123456789abcdef',
+                                   name=Transaction.ADD_ID, timestamp=datetime_utcnow())
```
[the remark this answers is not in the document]
mafesan: See my comment above: https://github.com/chaoss/grimoirelab-sortinghat/pull/218#discussion_r335912753 .

mafesan:
Hi @sduenas, I have just uploaded the changes so the PR is ready for your review again. Thanks!

tests/test_db.py
```diff
@@ -339,39 +349,77 @@ def test_name_none(self):
...
+
+        operations = Operation.objects.filter(trx=trx)
+        self.assertEqual(len(operations), 1)
+
+        op1 = operations[0]
+        self.assertIsInstance(op1, Operation)
+        self.assertEqual(op1.op_type, OpTypeChoice.ADD.value)
+        self.assertEqual(op1.entity_type, 'organization')
+        self.assertEqual(op1.trx, trx)
+        self.assertGreater(op1.timestamp, timestamp)
```
sduenas: What about checking `target` field? You should check that in every test.

sortinghat/core/api.py
```diff
@@ -152,23 +153,27 @@ def add_identity(source, name=None, email=None, username=None, uuid=None):
...
     except ValueError as e:
         raise InvalidValueError(msg=str(e))
 
     if not uuid:
-        uidentity = add_unique_identity_db(id_)
+        uidentity = add_unique_identity_db(trx, id_)
     else:
         uidentity = find_unique_identity(uuid)
 
+    args = {'trx': trx, 'uidentity': uidentity, 'identity_id': id_, 'source': source,
```
sduenas: Please format dicts as I commented before.

tests/test_api.py
```diff
@@ -383,6 +391,10 @@ def test_non_existing_uuid(self):
                              username=None,
                              uuid='FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF')
 
+        # Check if there are no transactions created when there is an error
+        transactions = Transaction.objects.all()
+        self.assertEqual(len(transactions), 2)
+
```
sduenas: This is not clear. There are already transactinos there but the number seems to be random. I think we should check this using the creation date, so you can check how many transactions were created after the API call. The same for the rest of the tests.

sortinghat/core/db.py
```diff
@@ -537,6 +631,13 @@ def move_identity(identity, uidentity):
     old_uidentity.save()
     uidentity.save()
 
+    trx.log_operation(op_type=OpTypeChoice.UPDATE, entity_type='identity',
+                      timestamp=datetime_utcnow(), args=input_args,
+                      target=input_args['identity'])
+    trx.log_operation(op_type=OpTypeChoice.UPDATE, entity_type='unique_identity',
+                      timestamp=datetime_utcnow(), args=input_args,
+                      target=input_args['uidentity'])
+
```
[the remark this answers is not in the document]
mafesan: We discussed about this and we agreed on not adding the log for this operation and deleting the log corresponding to the unique identity, as it is implicitly updated together with the `old_uidentity`.

tests/test_model.py
```diff
@@ -598,3 +602,90 @@ def test_last_modified(self):
...
+        self.assertLessEqual(trx.created_at, after_dt)
+
+        trx.save()
+
+        # Check if creation date does not change after saving the object
+        self.assertGreaterEqual(trx.created_at, before_dt)
+        self.assertLessEqual(trx.created_at, after_dt)
+
+
+class TestOperation(TransactionTestCase):
```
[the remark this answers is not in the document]
mafesan: I'm having difficulties testing this case, because this field is a `CharField` from Django models, whose `choices` options are the values from the enum class `OpType`.

When creating a `Operation` object directly using the call from Django models, it does not raise any error when the field `op_type` receives a explicit OpType value like `ADD` (string) or another string which is not in OpType (for instance, `test`). Any idea to get this under control?

Testing this case in `test_log.py` was not a problem, because the `log_operation` method now controls when `op_type` field is not an instance of `Operation.OpType`.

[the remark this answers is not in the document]
mafesan: It seems this check would be difficult to perform. After checking [the doc about Django choices](https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices) for `CharField`, it says that "If choices are given, they’re enforced by model validation". However, Django does not check this implicitly.

It does not check if the field's value is one of the choices or even if it is an integer instead of a string, but I added a test checking when `op_type` is `None` as this case is contemplated.

One of the alternatives would  be to declare a `clean()` method validating the field, but it would have to be called every time the object is created. 

As we are already checking this case when the `Operation` object is created in the `log_operation` method from `TransactionsLog`, we decided to decline this test.

sortinghat/core/log.py
```diff
@@ -0,0 +1,152 @@
...
+
+        :param name: Name of the method opening the transaction
+
+        :returns: a new TransactionsLog object containing the generated Transaction object
+        """
+        validate_field('name', name)
+
+        tuid = uuid.uuid4().hex
+
+        trx = Transaction(tuid=tuid, name=name, created_at=datetime_utcnow())
```
sduenas: Instead of calling to `Transaction` call to `cls`. This will make possible to inherit from `Transaction`.

sortinghat/core/log.py
```diff
@@ -0,0 +1,152 @@
...
+        :param timestamp: Datetime when the operation is created
+        :param args: Input arguments from the method creating the operation
+        :param target: Argument which the operation is directed to
+
+        :raises ClosedTransactionError: When trying to log an operation on a closed transaction
+        :raises TypeError: When the `op_type` is not an instance of `Operation.OpType` class
+
+        :returns: a new Operation object
+        """
+
```
sduenas: Remove this blank line.

sortinghat/core/errors.py
```diff
@@ -70,3 +72,10 @@ class InvalidValueError(BaseError):
 
     code = CODE_VALUE_ERROR
     message = "%(msg)s"
+
+
+class ClosedTransactionError(BaseError):
```
sduenas: Please move this commit before the previous one because it depends on it.

sortinghat/core/db.py
```diff
@@ -470,6 +544,10 @@ def add_enrollment(uidentity, organization,
         when `start < MIN_PERIOD_DATE`; or `end > MAX_PERIOD_DATE`
         or `start > end`.
     """
+    input_args = {'uidentity': uidentity.uuid,
+                  'organization': organization.name,
+                  'start': copy.deepcopy(str(start)), 'end': copy.deepcopy(str(end))}
+
```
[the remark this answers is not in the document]
mafesan: +1 to the formatting and to the renaming!

About keeping the arguments closer to the call, most of the times it is not possible because they are modified in the method, that is why they are saved in the beginning of the method. i have added a comment in these cases to make this clearer.

sortinghat/core/log.py
sduenas: My mistake, this is correctly done.

mafesan:
> All the code looks good but you need to add more content:
> 
>     * Add the dependency of django-mysql to `requirements.txt`
> 
>     * Add the new migration. Instead of removing existing migrations, you need file running
>       `./manage.py makemigrations --settings=config.settings.devel -n log core`
>       Use the flag ` --dry-run` before creating the file to check the result is like this one:
> 
> 
> ```
> Migrations for 'core':
>   sortinghat/core/migrations/0002_log.py
>     - Create model Operation
>     - Create model Transaction
>     - Change Meta options on enrollment
>     - Alter field end on enrollment
>     - Alter field start on enrollment
>     - Add field trx to operation
> ```

Oops, my bad @sduenas ! I have just submitted the changes addressing these missing files. Thanks!

PR #225, opened 2019-11-22
[schema] Add pagination filters for queries

mafesan:
Pagination has been implemented on top of the `Paginator` object from Django and applied to the existing queries.

Now, the query asking for organizations supports filtering by name.

When performing a query, there are two new parameters:
* `pageSize`: Number of elements we want to obtain per page. If `pageSize` value is not set, the value will be set using a configuration parameter defined in Django settings, named `DEFAULT_GRAPHQL_PAGE_SIZE` (currently set to `10` in `testing.py` configuration).
* `page`: Number of the page we are asking for. If `page` value is not set, it will return the first page
 by default.

These changes modify the structure of the queries. Now, the returned objects will be given under `entities`, and the information related with the pagination under `pageInfo`. For instance, to ask for a list of unique identities using pagination, the query would look like this:

```
{
  uidentities(
    page: 2
    pageSize: 5
  ){
    entities{
      uuid
    }
    pageInfo{
      page
      pageSize
      numPages
      hasNext
      hasPrev
      startIndex
      endIndex
      totalResults
    }
  }
}
```


tests/test_schema.py
```diff
@@ -624,9 +1063,45 @@ def test_filter_non_existing_registry(self):
                                                      self.timestamp.isoformat())
         executed = client.execute(test_query)
 
-        transactions = executed['data']['transactions']
+        transactions = executed['data']['transactions']['entities']
         self.assertListEqual(transactions, [])
 
+    def test_pagination(self):
+        """Check whether it returns the transactions searched when using pagination"""
+        timestamp = datetime_utcnow()
```
sduenas: Leave an empty line in between this two.

sortinghat/core/schema.py
```diff
@@ -121,6 +135,10 @@ class ProfileInputType(graphene.InputObjectType):
     country_code = graphene.String(required=False)
 
 
+class OrganizationFilterType(graphene.InputObjectType):
```
sduenas: This filter should go in a different PR.

tests/test_schema.py
mafesan: Done!

sortinghat/core/schema.py
mafesan: OK, I am already preparing the PR with these changes.

Issue #228, opened 2019-12-13
Unify omits blacklist when using email algorithm with `--no-strict-matching`


sanacl:
Having a Sortinghat database with blacklisted strings (emails) and executing the unify method with the matching algorithm `email` and `--no-strict-matching` produces and unexpected result. The unified identities do not use the blacklist so the identities with those values are merged.

How to reproduce:
* having a SH database add some emails to the blacklist with identities that are not merged
* merge by email with `--no-strict-matching`
* the result will be unified identities even when the email is present in the blacklist

sduenas:
The error is `email.py` does not check the blacklist when `--no-strict-matching` is active. This is the missing code, that is avaliable in `email_name.py`.

https://github.com/chaoss/grimoirelab-sortinghat/blob/master/sortinghat/matching/email_name.py#L142
https://github.com/chaoss/grimoirelab-sortinghat/blob/master/sortinghat/matching/email_name.py#L177

PR #229, opened 2019-12-13
[matching-email] Support blacklist when strict matching is False

valeriocos:
This code addresses https://github.com/chaoss/grimoirelab-sortinghat/issues/228, thus refactors the code to support the blacklisting of emails when the strict mode is set to `False`. The method `_check_blacklist` is created and the method `check_email` is modified accordingly (by removing the logic that blacklists emails).

Tests have been added accordingly.

Furthermore this PR removes the method `_filter_emails` which isn't used either in sortinghat or in other grimoirelab components.

### DIFF
--- a/sortinghat/matching/email.py
+++ b/sortinghat/matching/email.py
@@ -130,7 +130,7 @@ class EmailMatcher(IdentityMatcher):
         if fa.uuid and fb.uuid and fa.uuid == fb.uuid:
             return True
 
-        if fa.email in self.blacklist:
+        if self._check_blacklist(fa):
             return False
 
         # Compare email addresses first
@@ -160,6 +160,9 @@ class EmailMatcher(IdentityMatcher):
             if self.sources and id_.source.lower() not in self.sources:
                 continue
 
+            if self._check_blacklist(id_):
+                continue
+
             if self.strict:
                 if self._check_email(id_.email):
                     email = id_.email.lower()
@@ -180,13 +183,23 @@ class EmailMatcher(IdentityMatcher):
         """
         return ['email']
 
+    # not used
     def _filter_emails(self, ids):
         return [id_.email.lower() for id_ in ids
                 if self._check_email(id_.email)]
 
+    def _check_blacklist(self, id_):
+        if not id_.email:
+            return False
+
+        blacklisted = id_.email.lower() in self.blacklist
+
+        return blacklisted
+
     def _check_email(self, email):
         if not email:
             return False
-        elif email.lower() in self.blacklist:
-            return False
-        return self.email_pattern.match(email) is not None
+
+        checked = self.email_pattern.match(email) is not None
+
+        return checked
--- a/sortinghat/matching/email.py
+++ b/sortinghat/matching/email.py
@@ -183,11 +183,6 @@ class EmailMatcher(IdentityMatcher):
         """
         return ['email']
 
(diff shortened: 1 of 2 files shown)


tests/test_matcher_email.py
```diff
@@ -165,6 +165,85 @@ def test_match_with_blacklist(self):
...
+        jrae = UniqueIdentity(uuid='jrae')
+        jrae.identities = [Identity(name='Jane Rae', source='scm', uuid='jrae'),
+                           Identity(name='Jane Rae Doe', email='jane.rae@example.net', source='mls', uuid='jrae'),
+                           Identity(name='jrae', source='scm', uuid='jrae'),
+                           Identity(email='JRAE@example.net', source='scm', uuid='jrae')]
+
+        jane_rae = UniqueIdentity(uuid='Jane Rae')
+        jane_rae.identities = [Identity(name='Jane Rae', source='scm', uuid='Jane Rae'),
+                               Identity(email='jrae@example.net', source='mls', uuid='Jane Rae')]
+
```
sduenas: If the strict mode is `False` there should be at least one example where the field `email` is not a strict email address.

tests/test_matcher_email.py
```diff
@@ -424,6 +503,42 @@ def test_filter_identities_with_blacklist(self):
...
+                             Identity(username='jsmith', source='scm', uuid='jsmith'),
+                             Identity(email='', source='scm', uuid='jsmith')]
+
+        jrae = UniqueIdentity(uuid='jrae')
+        jrae.identities = [Identity(name='Jane Rae', source='scm', uuid='jrae'),
+                           Identity(name='Jane Rae Doe', email='jane.rae@example.net', source='mls', uuid='jrae'),
+                           Identity(name='jrae', source='scm', uuid='jrae'),
+                           Identity(email='JRAE@example.net', source='scm', uuid='jrae')]
+
+        bl = [MatchingBlacklist(excluded='jrae@example.net')]
```
sduenas: Same comment as in the test from above.

sortinghat/matching/email.py
```diff
@@ -180,13 +183,12 @@ def matching_criteria():
         """
         return ['email']
 
-    def _filter_emails(self, ids):
-        return [id_.email.lower() for id_ in ids
-                if self._check_email(id_.email)]
+    def _check_blacklist(self, id_):
+        blacklisted = False if not id_.email else id_.email.lower() in self.blacklist
+
+        return blacklisted
```
sduenas: This is easier to understand with `if..else` blocks.

sortinghat/matching/email.py
```diff
@@ -180,13 +183,12 @@ def matching_criteria():
...
 
     def _check_email(self, email):
-        if not email:
-            return False
-        elif email.lower() in self.blacklist:
-            return False
-        return self.email_pattern.match(email) is not None
+        checked = False if not email else self.email_pattern.match(email) is not None
```
sduenas: Same as above.

sortinghat/matching/email.py
```diff
@@ -180,13 +183,12 @@ def matching_criteria():
...
+
+        return checked
```
sduenas: Please leave an empty blank line at the end of the file.

valeriocos:
please @sduenas check it again

PR #232, opened 2019-12-23
[database] Allow SSL connections

### LINKED ISSUE #231: Allow SSL connections
When SortingHat uses MySQL or MariaDB server that requires SSL connections, the program raises an exception.

sduenas:
The engine was not configure to accept SSL connections. With this patch, the database module will try to establish a SSL connection first. If this is not possible, it will try to communicate with the server with a non-secure connection.

To set the SSL connection, it was necessary a minor hack that consists on passing an non-empty dictionary with key 'ssl' in 'connect_args'. The module pymysql expects a dict with some SSL paramters, which will force to use specific cert o ca. As we do not want that, we found that passing
a dict with a random key is valid and pymsql ignores that key but enforces a SSL connection.

Take into account this commit does not include tests. Although we want to have tests for every commit, I think this might be an exception due to `database` module does not have tests and this module will be replaced very soon by the new version available in `muggle` branch. On this branch, we use Django, so SSL will be configured and managed by that package.

If you want to test this patch, I recommend to create a database to force SSL connections. In a normal case, this wouldn't have worked. You can compare with a version without applying this patch that will raise an exception.

To do force SSL, just run:
```
GRANT ALL PRIVILEGES ON <database>.* TO <user>@'%' IDENTIFIED BY <password> REQUIRE SSL;
```

SortingHat should also work with no secure connections too.

This PR fixes #231 .

### DIFF
--- a/sortinghat/db/database.py
+++ b/sortinghat/db/database.py
@@ -106,8 +106,11 @@ class Database(object):
 
     @classmethod
     def build_engine(cls, user, password, database, host='localhost', port='3306'):
-        return create_database_engine(user, password, database,
-                                      host, port)
+        try:
+            return create_database_engine(user, password, database,
+                                          host, port)
+        except OperationalError as e:
+            raise DatabaseError(error=e.orig.args[1], code=e.orig.args[0])
 
     @classmethod
     def handle_database_error(cls, session, exception):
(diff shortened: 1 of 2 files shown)


mafesan:
I have tested these changes and I confirm it works as expected. 

* This is the execution on `master` branch:
​
```shell
$ ./sortinghat -u <user> -p <password> --host 127.0.0.1 orgs -a "Example"
(...)
sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (3159, 'Connections using insecure transport are prohibited while --require_secure_transport=ON.')
(Background on this error at: http://sqlalche.me/e/2j85)
```
* The same execution on `ssl-conn` branch did not returned any errors and added the organization to the database.

---
​
I had several problems trying to test this feature, so I am detailing some relevant notes (I have tested it using Ubuntu 18.04 LTS):

* When connecting to `mysql`, it is necessary to use the parameter `-h 127.0.0.1` to force the connection to be over TCP/IP protocol, so it will use SSL. Without this parameter, the connection will be established via UNIX socket and SSL connections will not be enabled.

```shell
$ mysql -u <user> -p -h 127.0.0.1
```
* Next step was to create the database and force the SSL connection (note the `'` surrounding the user and the password fields):
​
```SQL
GRANT ALL PRIVILEGES ON <database>.* TO '<user>'@'%' IDENTIFIED BY '<password>' REQUIRE SSL;
```
​
* And check that SSL is enabled:
​
```
mysql> status
-------------------
mysql  Ver 14.14 Distrib 5.7.28, for Linux (x86_64) using  EditLine wrapper
​
Connection id:		11
Current database:	test_pr
Current user:		oper@localhost
SSL:			Cipher in use is ECDHE-RSA-AES128-GCM-SHA256
(...)
Server version:		5.7.28-0ubuntu0.18.04.4 (Ubuntu)
Protocol version:	10
Connection:		127.0.0.1 via TCP/IP
```

* To force all the connections to use SSL, edit the file `/etc/mysql/my.cnf` adding these lines:

```
[mysqld]
require_secure_transport = ON
```

And then restart MySQL service: 
```shell
$ sudo systemctl restart mysql
```

Issue #233, opened 2019-12-27
[muggle] Perform authenticated GraphQL queries in Graphene-Django


mafesan:
The goal of this issue is to explore how to perform authenticated GraphQL queries in Graphene-Django, which are the different levels of permissions and how to apply them to a SortingHat database.

mafesan:
I have submitted a PR adding support for performing authenticated queries which is now ready for review: https://github.com/chaoss/grimoirelab-sortinghat/pull/234. Thanks!

sduenas:
Fixed in #234 .

PR #234, opened 2020-01-17
[schema] Add support for authenticated queries

mafesan:
This PR is related with issue #233.

## Summary

These changes add support for performing authenticated queries (both GraphQL queries and mutations), based on JSON Web Tokens (JWT).

To perform an authenticated query, an existing user must generate a token which has to be included in the `Authorization` header with the HTTP request. This token is generated using a mutation which comes defined by the `graphene-jwt` module, together with other extra mutations for verifying an existing token and refreshing it in case the token expired.

By default, all the queries and mutations now require authentication. This is controlled by a decorator named `check_auth` which comes from a pre-defined decorator named `login_required` and extending it to check both if the `user` from the `context.info` Django object is authenticated and the value of a new control variable named `AUTHENTICATION_REQUIRED` defined in Django settings allowing to set whether authentication is required or not.

If the user is not logged or uses an invalid token, the following exception is raised:

```
graphql.error.located_error.GraphQLLocatedError: You do not have permission to perform this action
```

Note that in Graphene-Django, all the exceptions raised which are caused by a GraphQL execution returns a result detailing the errors:

```
{
  "errors": [
    {
      "message": "Organization 'Test' already exists in the registry",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "addOrganization"
      ]
    }
  ],
  "data": {
    "addOrganization": null
  }
}
```

However, even if the exception is properly captured, a Traceback is printed to `STDERR` where the `manage.py` server is executed:

```
Traceback (most recent call last):
  File "/home/mafesan/devel/venvs/sortinghat_muggle/lib/python3.6/site-packages/promise/promise.py", line 487, in _resolve_from_executor
    executor(resolve, reject)
  File "/home/mafesan/devel/venvs/sortinghat_muggle/lib/python3.6/site-packages/promise/promise.py", line 754, in executor
    return resolve(f(*args, **kwargs))
  File "/home/mafesan/devel/venvs/sortinghat_muggle/lib/python3.6/site-packages/graphql/execution/middleware.py", line 75, in make_it_promise
    return next(*args, **kwargs)
  File "/home/mafesan/devel/venvs/sortinghat_muggle/lib/python3.6/site-packages/graphql_jwt/decorators.py", line 30, in wrapper
    return func(info.context, *args, **kwargs)
  File "/home/mafesan/devel/venvs/sortinghat_muggle/lib/python3.6/site-packages/graphql_jwt/decorators.py", line 41, in wrapper
    return f(*args, **kwargs)
  File "/home/mafesan/devel/grimoirelab-sortinghat/sortinghat/core/schema.py", line 220, in mutate
    org = add_organization(name)
  File "/usr/lib/python3.6/contextlib.py", line 52, in inner
    return func(*args, **kwds)
  File "/home/mafesan/devel/grimoirelab-sortinghat/sortinghat/core/api.py", line 378, in add_organization
    raise exc
  File "/home/mafesan/devel/grimoirelab-sortinghat/sortinghat/core/api.py", line 374, in add_organization
    org = add_organization_db(trxl, name=name)
  File "/home/mafesan/devel/grimoirelab-sortinghat/sortinghat/core/db.py", line 191, in add_organization
    _handle_integrity_error(Organization, exc)
  File "/home/mafesan/devel/grimoirelab-sortinghat/sortinghat/core/db.py", line 693, in _handle_integrity_error
    raise AlreadyExistsError(entity=entity, eid=eid)
graphql.error.located_error.GraphQLLocatedError: Organization 'Test' already exists in the registry
```

After researching about this behavior, it seems it is a known issue. There are several workarounds proposed, but there is not an official solution for this problem yet:

* This user is describing the same behavior: https://github.com/graphql-python/graphene-django/issues/780.
* This other user opened an older issue (https://github.com/graphql-python/graphene-django/issues/735) where the
proposed work around is to avoid this Traceback to be displayed by disabling the logging object which is printing it.
* [This user from Stack Overflow](https://stackoverflow.com/a/52881442) proposes a similar workaround for this behavior, limiting the depth of the Traceback.

We decided to go for this workaround, as the Traceback is not displayed anymore but the query result
is still including the information from `results.errors`, including what caused the error or exception when executing the query (for instance, adding an existing organization, not having permissions for performing some action, etc.).

## Configuration

There is a new package dependency: `graphql_jwt` module:

```
pip install django-graphql-jwt
```

In the Django settings script, it is necessary to include the following lines: 

```

AUTHENTICATION_BACKENDS = [
    'graphql_jwt.backends.JSONWebTokenBackend',
    'django.contrib.auth.backends.ModelBackend',
]

GRAPHENE = {
    'SCHEMA': 'sortinghat.core.schema',
    'MIDDLEWARE': [
        'graphql_jwt.middleware.JSONWebTokenMiddleware',
    ],
}

AUTHENTICATION_REQUIRED = True

```

### Unit tests

Note: After these changes, the `AUTHENTICATION_REQUIRED` variable in the testing configuration script must be set to `True`.

## About how to test these changes

To test Authentication, it is recommended to use _Insomnia_ or other application capable of modifying HTTP headers and cookie settings. As the GraphQL requests are sent via POST, there is an additional procedure to follow to configure the CSRF token for Django. These are the steps to follow:

* Go to `Headers` tab and set a new header named `X-CSRFToken` and for the value select the option "`Request -> Cookie`".

csrf1

* Once the "`Request -> Cookie`" option becomes a tag, click on it to edit it and fill the "Cookie name" field with `csrftoken`.

csrf2

Now we are ready for testing Authentication. First, we need to create a new user using the Django Admin interface in case we don't have it already. Then, we will use these mutations coming by default with `graphql_jwt` module: `tokenAuth`, `verifyToken` and `refreshToken`.

* `tokenAuth`: With this mutation we generate the token we must include in the HTTP headers to perform the authenticated requests.

```
mutation {
  tokenAuth(username: "user_test", password: "123456A!") {
    token
  }
}
```

* Once we have obtained the token after executing the last mutation, we will have to set the `Authentication` header and paste the token as value, preceded by `JWT ` (see next image).

```
{
  "data": {
    "tokenAuth": {
      "token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6Im1hZeHAiOjE1Ngz5MTAmVzYW4iLCJlzMzYsIm9yaWdJYXQiOjE1Nzg5MTAwMzZ9.wUf1P02aUfWqC-my7YwBOhM7awlk5KaoqAfTKNIx0mk"
    }
  }
}
```

jwt

* We can use the `verifyToken` mutation to check if the token is valid:

```
mutation {
  verifyToken(token: "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VybmFtZSI6Im1hZeHAiOjE1Ngz5MTAmVzYW4iLCJlzMzYsIm9yaWdJYXQiOjE1Nzg5MTAwMzZ9.wUf1P02aUfWqC-my7YwBOhM7awlk5KaoqAfTKNIx0mk") {
    payload
  }
}
```

* And the `refreshToken` mutation to obtain a new token:

```
mutation RefreshToken($token: String!) {
  refreshToken(token: $token) {
    token
    payload
  }
}
```

### DIFF
--- a/sortinghat/core/decorators.py
+++ b/sortinghat/core/decorators.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014-2019 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#     Santiago Dueñas <sduenas@bitergia.com>
+#     Miguel Ángel Fernández <mafesan@bitergia.com>
+#
+
+from django.conf import settings
+from graphql_jwt.decorators import user_passes_test
+
+
+# This custom decorator takes the `user` object from the request's
+# context and checks the value of the `is_authenticated` variable
+# and the `AUTHENTICATION_REQUIRED` variable from the config settings.
+check_auth = user_passes_test(
+    lambda u: u.is_authenticated or not settings.AUTHENTICATION_REQUIRED
+)
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -20,8 +20,10 @@
 #     Miguel Ángel Fernández <mafesan@bitergia.com>
 #
 
-import graphene
 import json
+
+import graphene
+import graphql_jwt
 
 from django.conf import settings
 from django.core.paginator import Paginator
@@ -33,6 +35,7 @@ from graphene.types.generic import GenericScalar
 from graphene_django.converter import convert_django_field
 from graphene_django.types import DjangoObjectType
 
+from .decorators import check_auth
 from .api import (add_identity,
                   delete_identity,
                   update_profile,
@@ -210,6 +213,7 @@ class AddOrganization(graphene.Mutation):
 
     organization = graphene.Field(lambda: OrganizationType)
 
+    @check_auth
     def mutate(self, info, name):
         org = add_organization(name)
 
@@ -224,6 +228,7 @@ class DeleteOrganization(graphene.Mutation):
 
     organization = graphene.Field(lambda: OrganizationType)
 
+    @check_auth
     def mutate(self, info, name):
         org = delete_organization(name)
 
@@ -240,6 +245,7 @@ class AddDomain(graphene.Mutation):
 
     domain = graphene.Field(lambda: DomainType)
 
+    @check_auth
     def mutate(self, info, organization, domain, is_top_domain=False):
         dom = add_domain(organization, domain, is_top_domain=is_top_domain)
 
@@ -254,6 +260,7 @@ class DeleteDomain(graphene.Mutation):
 
     domain = graphene.Field(lambda: DomainType)
 
+    @check_auth
     def mutate(self, info, domain):
         dom = delete_domain(domain)
 
@@ -273,6 +280,7 @@ class AddIdentity(graphene.Mutation):
     uuid = graphene.Field(lambda: graphene.String)
     uidentity = graphene.Field(lambda: UniqueIdentityType)
 
+    @check_auth
     def mutate(self, info, source,
                name=None, email=None, username=None,
                uuid=None):
@@ -298,6 +306,7 @@ class DeleteIdentity(graphene.Mutation):
     uuid = graphene.Field(lambda: graphene.String)
     uidentity = graphene.Field(lambda: UniqueIdentityType)
 
+    @check_auth
     def mutate(self, info, uuid):
         uidentity = delete_identity(uuid)
 
@@ -315,6 +324,7 @@ class UpdateProfile(graphene.Mutation):
     uuid = graphene.Field(lambda: graphene.String)
     uidentity = graphene.Field(lambda: UniqueIdentityType)
 
+    @check_auth
     def mutate(self, info, uuid, data):
         uidentity = update_profile(uuid, **data)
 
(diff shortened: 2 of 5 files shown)


sortinghat/core/decorators.py
```diff
@@ -0,0 +1,28 @@
...
+#
+# Authors:
+#     Santiago Dueñas <sduenas@bitergia.com>
+#     Miguel Ángel Fernández <mafesan@bitergia.com>
+#
+
+from django.conf import settings
+from graphql_jwt.decorators import user_passes_test
+
+auth_required = user_passes_test(
```
sduenas: I think we should rename this function to something like `check_auth` or similar. I'm saying this because the name does something different that the function really does. If `settings.AUTHENTICATION_REQUIRED` is deactivated, the function evaluates to True, so you are saying that authentication is not required but when this is active, you are checking if the user is authenticated.

I also think we should document this somehow or we are going to forget why we're doing this this way. I think it will be enough if you define this as:

```
def auth_required:
    return user_passes_test(
        lambda u: u.is_authenticated or not settings.AUTHENTICATION_REQUIRED
    )
```
Inside the function you can explain that this works as a decorator and how to use it.

sortinghat/core/schema.py
```diff
@@ -21,6 +21,8 @@
 #
 
 import graphene
+import graphql_jwt
+
 import json
```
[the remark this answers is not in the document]
mafesan: Thank you for the heads-up! Done.

sortinghat/core/schema.py
```diff
@@ -506,3 +524,7 @@ class SortingHatMutation(graphene.ObjectType):
     merge_identities = MergeIdentities.Field()
     enroll = Enroll.Field()
     withdraw = Withdraw.Field()
+    # JWT authentication
```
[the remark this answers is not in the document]
mafesan: Done.

sortinghat/core/decorators.py
mafesan: You are right, I have renamed `auth_required` to `check_auth`. Moreover, I have added a comment explaining briefly what this decorator does.

mafesan: About defining the decorator as proposed, I was having difficulties defining it as a method because it received the calling method as positional argument.

In this case, the method would have to be defined as:

```python
def check_auth(func):
    return user_passes_test(
        lambda u: u.is_authenticated or not settings.AUTHENTICATION_REQUIRED
    )(func)
```
@sduenas and me have agreed on keeping this decorator as it was for the moment, but adding the comments I mentioned above.

mafesan:
Done! Please @sduenas check if I addressed you comments as expected. Thanks!

PR #235, opened 2020-01-22
[setup] Let pandas be >= 0.22.0

jgbarah:
If we pin pandas at 0.22.0, it doesn't build with Python 3.7.x, apparently because of problems when building numpy for pandas.

Letting pandas be greater or equal than 0.22.0 doesn't seem to break anything, since platforms having only 0.22.0, as Travis had in July'19 (see #206), would still fulfill the dependency.

Signed-off-by: Jesus M. Gonzalez-Barahona <jgb@gsyc.es>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ setup(name="sortinghat",
         'sqlalchemy>=1.2',
         'jinja2',
         'python-dateutil>=2.6.0',
-        'pandas==0.22.0',
+        'pandas>=0.22.0,<=0.25.3',
         'pyyaml>=3.12',
         'requests>=2.9',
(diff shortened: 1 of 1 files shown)


jgbarah:
> OK from my side. I guess also this dependency should be updated: https://github.com/chaoss/grimoirelab-elk/blob/master/requirements.txt#L7, right?

Yes. I was waiting to see what happens with this one ;-) Once it is merged, I plan to go with the other.

(In fact, we can start with this one, since the change is, I think, compatible with the current situation in grimoire-elk, and then do the same change there)

valeriocos:
Ok! WDYT @sduenas ?

Issue #236, opened 2020-01-24
[muggle] Modify merge identities method to accept a list as input


mafesan:
The method in charge of merging identities should accept a list of identities as input, so multiple identities can be merged at once. This operation must be atomic, so we may need to check if this causes any consistency errors.

PR #240, opened 2020-02-06
[muggle] Add support for merging multiple identities at once

mafesan:
The `merge_identities` mutation has been modified to accept a list of unique identities in the `from_uuid` field (now renamed to `from_uuids`). 

The API method for merging identities has been adapted to support this list as an input. New sub-methods have been included, iterating over the set of unique identities to merge.

Note: This PR is related to issue #236 .

### DIFF
--- a/sortinghat/core/api.py
+++ b/sortinghat/core/api.py
@@ -45,7 +45,7 @@ from .db import (find_unique_identity,
                  delete_enrollment)
 from .errors import InvalidValueError, AlreadyExistsError, NotFoundError
 from .log import TransactionsLog
-from .models import MIN_PERIOD_DATE, MAX_PERIOD_DATE
+from .models import Identity, MIN_PERIOD_DATE, MAX_PERIOD_DATE
 from .utils import unaccent_string, merge_datetime_ranges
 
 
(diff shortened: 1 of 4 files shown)


sortinghat/core/api.py
```diff
@@ -735,21 +735,54 @@ def merge_identities(from_uuid, to_uuid):
...
+        `to_uuid` do not exist in the registry
     """
-    def _merge_enrollments(trxl, from_uid, to_uid):
-        """Merge enrollments from two `UniqueIdentity` objects"""
-        # Get current enrollments from both uidentities
-        enrollments_db = from_uid.enrollments.all() | to_uid.enrollments.all()
-        enrollments = {}
 
+    def _find_unique_identities(from_uuids, to_uuid):
+        """Find the unique identities to be merged from their uuids"""
```
mafesan: Question: Should I include a black space between this _docstring_ line and the next one? I am not sure if this the right way to do it when it is an inner method.

sduenas: Yep. You should.

mafesan: Done, thank you!

sortinghat/core/api.py
```diff
@@ -714,42 +714,84 @@ def withdraw(uuid, organization, from_date=None, to_date=None):
...
+            # Check whether input values are valid
+            if from_uuid is None:
+                raise InvalidValueError(msg="'from_uuid' cannot be None")
+            if from_uuid == '':
+                raise InvalidValueError(msg="'from_uuid' cannot be an empty string")
+            if from_uuid == to_uuid:
+                raise InvalidValueError(msg="'from_uuid' and 'to_uuid' cannot be equal")
+
+            try:
+                unique_identities.append(find_unique_identity(from_uuid))
```
sduenas: Would you mind to store the uidentity first in a variable and then append that variable to the list? I think that's more clear.

sortinghat/core/api.py
```diff
@@ -771,59 +813,60 @@ def _merge_enrollments(trxl, from_uid, to_uid):
...
-        if not to_uid.profile.gender_acc:
-            to_uid.profile.gender_acc = from_uid.profile.gender_acc
-        if not to_uid.profile.country:
-            to_uid.profile.country = from_uid.profile.country
+    def _merge_profiles(from_uids, to_uid):
+        """Merge the profiles from `UniqueIdentity` objects"""
+
+        for from_uid in from_uids:
+            if from_uid.profile.is_bot or to_uid.profile.is_bot:
+                to_uid.profile.is_bot = True
```
sduenas: Wouldn't be better to limit this to `if not to_uid.profile.is_bot and from_uid.profile.is_bot`?

sortinghat/core/api.py
```diff
@@ -771,59 +813,60 @@ def _merge_enrollments(trxl, from_uid, to_uid):
...
+            delete_unique_identity_db(trxl, uid)
+
     # Check input values
-    if from_uuid is None:
-        raise InvalidValueError(msg="'from_uuid' cannot be None")
-    if from_uuid == '':
-        raise InvalidValueError(msg="'from_uuid' cannot be an empty string")
+    if from_uuids is None:
+        raise InvalidValueError(msg="'from_uuids' cannot be None")
+    if from_uuids == []:
```
sduenas: It might be better to evaluate it with `len`.

sortinghat/core/api.py
```diff
@@ -771,59 +813,60 @@ def _merge_enrollments(trxl, from_uid, to_uid):
...
 
-    try:
-        identities = from_uid.identities.all()
-    except ValueError as e:
-        raise InvalidValueError(msg=str(e))
-
-    for identity in identities:
-        move_identity_db(trxl, identity, to_uid)
+    identities = _get_identities(from_uids)
+    _move_identities(trxl, identities, to_uid)
```
sduenas: Maybe it's better to call `_get_identities` inside `_move_identities` because `identities` variable is not used anywhere else. It also will match with the prototype of the other functions.

sortinghat/core/api.py
mafesan: You are totally right, now the condition is redundant. I'm changing this.

sortinghat/core/api.py
mafesan: The problem I see evaluating this with `len` is that both values `[]` and `''` would return `0` using `len`. That's why I set the check for an empty list here, and another check for an empty string when iterating over the elements of a non-empty list. 
It is also true that when the field `from_uuids` could be pre-validated, so we could assume we could receive an empty list as value but we couldn't receive an empty string, so I'm not sure if this check is redundant or not. Does any of this make sense to you? Thanks in advance!

sduenas: Ok! Makes sense then.

mafesan:
I have just addressed your comments @sduenas, please have a look. Thank you!

Issue #121, opened 2018-02-13
"Error: Duplicate CHECK constraint name 'is_bot_check'" when initializing database on MariaDB ≥10.2


aklapper:
```
$:acko\> mysql -uuser
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 18
Server version: 10.2.12-MariaDB MariaDB Server
MariaDB [(none)]> drop database sortinghat_wm;
Query OK, 4 rows affected (0.09 sec)
MariaDB [(none)]> drop database sortinghat_wm;
ERROR 1008 (HY000): Can't drop database 'sortinghat_wm'; database doesn't exist
MariaDB [(none)]> Bye
$:acko\> sortinghat init sortinghat_wm
Error: Duplicate CHECK constraint name 'is_bot_check' (err: 1826)
$:acko\> grep -r "is_bot_check" .
Binary file ./sortinghat/db/model.pyc matches
./sortinghat/db/model.py:    is_bot = Column(CoerceToBool(name='is_bot_check'), default=False)
./build/lib/sortinghat/db/model.py:    is_bot = Column(CoerceToBool(name='is_bot_check'), default=False)
$:acko\> git show
commit b6f8031174e8e82b7141ea9cfcddc75af9d91d04 (HEAD -> master, origin/master, origin/HEAD)
Merge: 84eceb3 0f66c9d
Author: Santiago Dueñas <sduenas@bitergia.com>
Date:   Tue Feb 13 17:38:11 2018 +0100
```

sduenas:
Weird.

@valeriocos can you check this please, it's related with https://github.com/chaoss/grimoirelab-sortinghat/commit/5f69ed899c94584de17d47b37152098f64012e10

valeriocos:
Hi @aklapper, 

I'm not able to replicate the issue you are reporting.  I followed the same steps (see below) using the native and docker installations.

```
:~/mysql -uroot
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 128
Server version: 10.0.33-MariaDB-0ubuntu0.16.04.1 Ubuntu 16.04

MariaDB [(none)]> drop database sortinghat_wm;
Query OK, 8 rows affected (0.01 sec)

MariaDB [(none)]> drop database sortinghat_wm;
ERROR 1008 (HY000): Can't drop database 'sortinghat_wm'; database doesn't exist

[...]

PR #249, opened 2020-02-21
[db-model] Remove CoerceToBool type

valeriocos:
This code fixes the bug #121, thus it removes the type CoerceToBool (introduced at 5f69ed8), which was used to handle string representations of boolean values (i.e., True and False) for SqlAlchemy versions prior to 1.1.14. This was due to the fact that SqlAlchemy versions 1.1.x defined the Boolean type as follows:
```
Boolean typically uses BOOLEAN or SMALLINT on the DDL side,
and on the Python side deals in True or False.
```

Thus, equivalent string representations weren't handled and required the definition of the CoerceToBool (which created an additional CHECK constraint).

From versions greater than >=1.2.0 (already set in the current setup.py),
the CoerceToBool isn't needed anymore and can be removed. In particular, the
Boolean type is now defined as:
```
The Boolean datatype currently has two levels of assertion that the values
persisted are simple true/false values. For all backends, only the Python
values None, True, False, 1 or 0 are accepted as parameter values. For those
backends that don’t support a “native boolean” datatype, a CHECK constraint
is also created on the target column. Production of the CHECK constraint
can be disabled by passing the Boolean.create_constraint flag set to False
```

Tests have been updated accordingly.

### DIFF
--- a/sortinghat/db/model.py
+++ b/sortinghat/db/model.py
@@ -29,7 +29,7 @@ from sqlalchemy.orm import backref, relationship
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.orderinglist import ordering_list
-from sqlalchemy.types import TypeDecorator, Boolean
+from sqlalchemy.types import Boolean
 
 # Default dates for periods
 MIN_PERIOD_DATE = datetime.datetime(1900, 1, 1, 0, 0, 0)
@@ -48,21 +48,6 @@ MAX_SIZE_CHAR_COLUMN = 191
 
 logger = logging.getLogger(__name__)
 ModelBase = declarative_base()
-
-
-class CoerceToBool(TypeDecorator):
-    impl = Boolean
-
-    def process_bind_param(self, value, dialect):
-        if type(value) == bool:
-            value = [False, True].index(value)
-        elif type(value) == int:
-            if value not in [0, 1]:
-                raise ValueError
-        else:
-            raise ValueError
-
-        return value
 
 
 class Organization(ModelBase):
@@ -95,7 +80,7 @@ class Domain(ModelBase):
 
     id = Column(Integer, primary_key=True)
     domain = Column(String(128), nullable=False)
-    is_top_domain = Column(CoerceToBool(name='top_domain_check'), default=False)
+    is_top_domain = Column(Boolean(name='top_domain_check'), default=False)
     organization_id = Column(Integer,
                              ForeignKey('organizations.id', ondelete='CASCADE',
                                         onupdate='CASCADE'),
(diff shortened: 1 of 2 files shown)


valeriocos:
> Why did CoerceToBool exist and what did it try to solve?

This is the commit message at https://github.com/chaoss/grimoirelab-sortinghat/commit/5f69ed899c94584de17d47b37152098f64012e10

```
[model] Coerce boolean type

This patch restricts the accepted values for booleans to True, False
0 and 1 using a sqlalchemy typedecorator. This fix is equivalent to
adding check constraints on boolean DB columns, however this kind
of constraint is ignored in MySQL DBs.
```

> Why was is causing problems?

This is the PR where this change has been introduced: https://github.com/chaoss/grimoirelab-sortinghat/pull/108

I guess that this change was needed due to the updates of sqlalchemy and pymysql versions. I didn't find any extra information in the other github issues or internal tickets (since you reviewed that PR, maybe you remember something). I tried with different combinations of sqlalchemy and pymysql and running the tests on different commits, but I wasn't able to reproduce the CHECK error.

> Why it's not needed anymore?

This PR shows that removing CoerceToBool doesn't break any test and allow to move to more recent version of MariaDb. Find below some of tests I did.

- Mariadb 10.0

   **Test 1**
   Master branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: No error about the CHECK constraint are thrown

   **Test 2**
   PR branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: No error about the CHECK constraint are thrown

- Mariadb 10.1

   **Test 1**
   Master branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: No error about the CHECK constraint are thrown

   **Test 2**
   PR branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: No error about the CHECK constraint are thrown. However the error type has changed

- Mariadb 10.2

   **Test 1**
   Master branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: errors similar to `sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (1826, "Duplicate CHECK constraint name 'top_domain_check'")`

  **Test 2**
   PR branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: No error about the CHECK constraint are thrown. However the error type has changed

- Mariadb 10.3

   **Test 1**
   Master branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: errors similar to `sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (1826, "Duplicate CHECK constraint name 'top_domain_check'")`

  **Test 2**
   PR branch
   PyMySQL>=0.7.0
   sqlalchemy>=1.2.0
   Result: No error about the CHECK constraint are thrown

Issue #237, opened 2020-01-24
[muggle] Add an atomic method to split identities


mafesan:
Similar to the method for merging identities, this method should also accept a list of identities as input, so multiple identities can be splitted at once. This operation must be atomic, so we may need to check if this causes any consistency errors.

This issue is related to https://github.com/chaoss/grimoirelab-sortinghat/issues/236.

mafesan:
I have submitted a PR for this feature, which is now ready for review: #250 . Thanks!

sduenas:
Fixed in #250.

PR #250, opened 2020-02-24
Add support for unmerging multiple identities at once

mafesan:
This method accepts a list of identities as input, so multiple identities can be unmerged/splitted at once, creating a unique identity for each one. 

`name` and `email` fields from every identity will be set to the corresponding profile fields when the new unique identity is created. Nor the enrollments or the profile from any parent unique identity of the input identities are modified.

When a given identity `uuid` is equal to the `uuid` from its parent unique identity, there will be no effect.

Note: This PR is related to issue #237.

### DIFF
--- a/sortinghat/core/api.py
+++ b/sortinghat/core/api.py
@@ -888,3 +888,95 @@ def merge_identities(ctx, from_uuids, to_uuid):
     trxl.close()
 
     return to_uid
+
+
+@django.db.transaction.atomic
+def unmerge_identities(ctx, uuids):
+    """
+    Unmerge one or more identities from their corresponding unique identities.
+
+    Use this function to separate a list of `uuid` identities, creating a unique
+    identity for each one. A profile for each new unique identity will be created using
+    the `name` and `email` fields of the corresponding identity.
+    Nor the enrollments or the profile from any parent unique identity of the input
+    identities are modified.
+
+    When a given identity `uuid` is equal to the `uuid` from its parent unique
+    identity, there will be no effect.
+
+    The function raises a `NotFoundError` exception when either any `uuid` from
+    the list does not exist in the registry.
+
+    The function raises an `InvalidValueError` exception when either any `uuid`
+    from the list is `None` or an empty string.
+
+    :param ctx: context from where this method is called
+    :param uuids: list of identifiers of the identities set to be unmerged
+
+    :returns: a list of UniqueIdentity objects as the result of the unmerged identities
+
+    :raises NotFoundError: raised when either any `uuid` from the list
+        does not exist in the registry
+    :raises InvalidValueError: raised when either any `uuid` from the
+        list is `None` or an empty string
+    """
+    def _find_identities(uuids):
+        """Find the identities to be unmerged from their parent unique identities"""
+
+        identities = []
+        for uuid in uuids:
+            # Check whether input values are valid
+            if uuid is None:
+                raise InvalidValueError(msg="'uuid' cannot be None")
+            if uuid == '':
+                raise InvalidValueError(msg="'uuid' cannot be an empty string")
+
+            uid = find_identity(uuid)
+            identities.append(uid)
+
+        return identities
+
+    def _set_destination_identity(trxl, identity):
+        """Create the unique identity, if it does not exist, where the identity will be moved to."""
+
+        if identity.id != identity.uidentity_id:
+            uid = add_unique_identity_db(trxl, identity.id)
+            uid = update_profile_db(trxl, uid,
+                                    name=identity.name,
+                                    email=identity.email)
+        else:
+            # Case when the identity to be unmerged is the main one
+            uid = identity.uidentity
+
+        return uid
+
+    def _move_to_destination(trxl, identity, uid):
+        """Move the identity from the old unique identity to the new one"""
+
+        try:
+            uidentity = move_identity_db(trxl, identity, uid)
+        except ValueError:
+            # Case when the identity is already assigned to the unique identity
+            uidentity = uid
+
+        return uidentity
+
+    # Check input value
+    if uuids is None:
+        raise InvalidValueError(msg="'uuids' cannot be None")
+    if uuids == []:
+        raise InvalidValueError(msg="'uuids' cannot be an empty list")
+
+    trxl = TransactionsLog.open('unmerge_identities', ctx)
+
+    identities = _find_identities(uuids)
+
+    new_uidentities = []
+    for identity in identities:
+        uid = _set_destination_identity(trxl, identity)
+        uidentity = _move_to_destination(trxl, identity, uid)
+        new_uidentities.append(uidentity)
+
+    trxl.close()
+
+    return new_uidentities
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -50,6 +50,7 @@ ALREADY_EXISTS_ERROR = "{entity} already exists in the registry"
 SOURCE_NONE_OR_EMPTY_ERROR = "'source' cannot be"
 IDENTITY_NONE_OR_EMPTY_ERROR = "identity data cannot be empty"
 UUID_NONE_OR_EMPTY_ERROR = "'uuid' cannot be"
+UUIDS_NONE_OR_EMPTY_ERROR = "'uuids' cannot be"
 FROM_ID_NONE_OR_EMPTY_ERROR = "'from_id' cannot be"
 FROM_UUID_NONE_OR_EMPTY_ERROR = "'from_uuid' cannot be"
 FROM_UUIDS_NONE_OR_EMPTY_ERROR = "'from_uuids' cannot be"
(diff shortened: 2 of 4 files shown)


sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+
+    The function raises an `InvalidValueError` exception when either any `uuid`
+    from the list is `None` or an empty string.
+
+    :param ctx: context from where this method is called
+    :param uuids: List of identifiers of the identities set to be unmerged
+
+    :returns: a list of UniqueIdentity objects as the result of the unmerged identities
+
+    :raises NotFoundError: raised when when either any `uuid` from the list
```
sduenas:
```suggestion
    :raises NotFoundError: raised when either any `uuid` from the list
```

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+        does not exist in the registry
+    :raises InvalidValueError: raised when when either any `uuid` from the
```
sduenas:
```suggestion
    :raises InvalidValueError: raised when either any `uuid` from the
```

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+        list is `None` or an empty string
+    """
+
```
sduenas: Remove this blank line.

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+            if uuid is None:
+                raise InvalidValueError(msg="'uuid' cannot be None")
+            if uuid == '':
+                raise InvalidValueError(msg="'uuid' cannot be an empty string")
+
+            try:
+                uid = find_identity(uuid)
+                identities.append(uid)
+            except NotFoundError as exc:
+                raise exc
```
sduenas: If we are not handling the exception, it's better to delete the `try...raise` block.

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+
+        return identities
+
+    def _set_destination_identity(trxl, identity):
+        """
+        Create the new unique identity where the identity
+        will be moved to if it does not exist
+        """
```
sduenas:
```suggestion
        """Create the unique identity, if it does not exist, where the identity will be moved to."""
```

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+                                    email=identity.email)
+        else:
+            # Case when the identity to be unmerged is the main one
+            uid = identity.uidentity
+
+        return uid
+
+    def _move_to_destination(trxl, identity, uid):
+        """Move the identity from the old unique identity to the new one"""
+        try:
```
sduenas: Leave a blank space in between these.

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+    if uuids == []:
+        raise InvalidValueError(msg="'uuids' cannot be an empty list")
+
+    trxl = TransactionsLog.open('unmerge_identities', ctx)
+
+    try:
+        identities = _find_identities(uuids)
+    except NotFoundError as exc:
+        # At least one identity was not found, so they cannot be unmerged
+        raise exc
```
sduenas: Same as before.

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+
+    new_uids = []
```
sduenas: It will be better to rename this to `new_uidentities`. It's more clear because `uids` looks like you are storing identifiers instead of identities.

tests/test_api.py
```diff
@@ -3659,3 +3660,357 @@ def test_operations(self):
...
+                   to_date=datetime.datetime(2017, 6, 1))
+
+        api.add_identity(self.ctx, 'scm', email='jsmith@bitergia')
+        api.add_identity(self.ctx,
+                         'phabricator',
+                         email='jsmith-phab@bitergia',
+                         uuid='e8284285566fdc1f41c8a22bb84a295fc3c4cbb3')
+        api.update_profile(self.ctx,
+                           uuid='e8284285566fdc1f41c8a22bb84a295fc3c4cbb3', name='John Smith',
+                           email='jsmith@profile-email', is_bot=True, country_code='US')
```
sduenas: The profile was already edited before. What's the purpose of this? Why don't you do it at the same time?

tests/test_api.py
```diff
@@ -3659,3 +3660,357 @@ def test_operations(self):
...
+                         'phabricator',
+                         email='jsmith2@libresoft',
+                         uuid='e8284285566fdc1f41c8a22bb84a295fc3c4cbb3')
+        api.add_identity(self.ctx,
+                         'phabricator',
+                         email='jsmith3@libresoft',
+                         uuid='e8284285566fdc1f41c8a22bb84a295fc3c4cbb3')
+        api.update_profile(self.ctx,
+                           uuid='e8284285566fdc1f41c8a22bb84a295fc3c4cbb3', name='John Smith',
+                           email='jsmith@profile-email', is_bot=False, country_code='US')
```
sduenas: Same question as before.

tests/test_api.py
```diff
@@ -3659,3 +3660,357 @@ def test_operations(self):
...
+        identities = uidentity.identities.all()
+        self.assertEqual(len(identities), 1)
+
+        id1 = identities[0]
+        self.assertEqual(id1.id, '67fc4f8a56aa12ab981d2a4c1de065bb9936c9f6')
+        self.assertEqual(id1.email, 'jsmith-git@example')
+        self.assertEqual(id1.source, 'git')
+
+        enrollments = uidentity.enrollments.all()
+        self.assertEqual(len(enrollments), 0)
```
sduenas: I think you should check here that the other identities weren't move from the original unique identity and that other things as the profile or the enrollments didn't change either.

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+
+
+@django.db.transaction.atomic
+def unmerge_identities(ctx, uuids):
+    """
+    Unmerge one or more identities from their corresponding unique identities.
+
+    Use this function to separate a list of `uuid` identities, creating a unique
+    identity for each one. `name` and `email` fields from every identity will be
+    set to the corresponding profile fields when the new unique identity is created.
```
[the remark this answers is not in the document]
mafesan: Done!

sortinghat/core/api.py
```diff
@@ -888,3 +888,104 @@ def _delete_unique_identities(trxl, uids):
...
+    identity, there will be no effect.
+
+    The function raises a `NotFoundError` exception when either any `uuid` from
+    the list does not exist in the registry.
+
+    The function raises an `InvalidValueError` exception when either any `uuid`
+    from the list is `None` or an empty string.
+
+    :param ctx: context from where this method is called
+    :param uuids: List of identifiers of the identities set to be unmerged
```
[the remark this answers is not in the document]
mafesan: Done!

sortinghat/core/api.py
mafesan: Oops! My bad :sweat_smile:

sortinghat/core/api.py
mafesan: Done

sortinghat/core/api.py
mafesan: Done!

sortinghat/core/api.py
mafesan: Done.

sortinghat/core/api.py
mafesan: Done.

sortinghat/core/api.py
mafesan: Done.

sortinghat/core/api.py
mafesan: Done.

sortinghat/core/api.py
mafesan: Done.

tests/test_api.py
mafesan: You are right, I removed the unnecessary calls to `update_profile`.

tests/test_api.py
mafesan: Done.

tests/test_api.py
mafesan: Done, I have checked the identities, profile and enrollment from the former parent unique identity.

mafesan:
@sduenas Thank you for your comments! I have just submitted the changes, please review.

Issue #239, opened 2020-01-24
[muggle] Implement command line/web client


mafesan:
The goal is to implement a command line client, mimicking the behavior from the standard version of SortingHat.
​
As we need to implement a web client too, we need to decide if we are implementing a client based on Apollo technology and then porting it to work as a command line client, or we are taking the alternative of creating a command line client which can be ported to work as a web client too.

PR #254, opened 2020-03-16
First implementation of a basic client

sduenas:
This commit adds a naive implementation of a SortingHat client. It allows to run any query or mutation supported by the server calling the method `execute()`.

This client uses `sgqlc` library that allows to define GraphQL operations faster.

Related to #239.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,11 @@
 Django==2.1
-grimoirelab_toolkit>=0.1.8
-mysqlclient==1.3.13
-python-dateutil>=2.8.0
-graphene-django>=2.0
 django-mysql>=3.2.0
 django-graphql-jwt>=0.3.0
+graphene-django>=2.0
+sgqlc>=10.1
+mysqlclient==1.3.13
+python-dateutil>=2.8.0
+grimoirelab_toolkit>=0.1.8
+requests>=2.7.0
+uri>=2.0.1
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,7 +1,10 @@
...
-graphene-django>=2.0
 django-mysql>=3.2.0
 django-graphql-jwt>=0.3.0
+graphene-django>=2.0
+sgqlc>=10.1
+mysqlclient==1.3.13
+python-dateutil>=2.8.0
+grimoirelab_toolkit>=0.1.8
+requests>=2.7.0
+uri>=2.0.1
```
mafesan: I think there is a missing requirement here, which is the `httpretty` module.

sortinghat/cli/client/client.py
```diff
@@ -0,0 +1,149 @@
...
+        return self.msg
+
+
+class SortingHatClient:
+    """SortingHat client.
+
+    This client allows to run operations in the SortingHat server
+    that listens in `host` and `port`.
+
+    After initializing an instance, call to `connect` to establish
```
mafesan: I think `call` should be replaced by `calls`.

sduenas: It's an imperative form in this case, so I think it's right. In any case, we can revisit the documentation after we make some progress.

PR #255, opened 2020-03-26
Add cmd command

### LINKED ISSUE #174: Redefine Organizations command line client
Since SortingHat is being redefined as a web service, the command line client does not work. The first step to move the client to use the server will be to support the `orgs` command.

The command line client should use the GraphQL API to perform this operation. Take into account `orgs` comand was able to list organizations, add/delete organizations and add/delete domains.

This tasks requires of #173.

sduenas:
This commands re-implements `orgs` command. The differences with the original command are:

* `orgs` doesn't list organizations by default, it should be run using `orgs show` instead
* `--term` parameter in `show` is not supported because the server doesn't support it either
* `--overwrite` parameter in `add` has been removed

Fixes #174


sduenas:
@mafesan, I've just pushed some new commits to fix the problems we found while testing this.

I've fixed the problems with the invalid token and how to handle connections and the errors raised.

sortinghat/cli/cmds/orgs.py
```diff
@@ -0,0 +1,240 @@
...
+#
+# Authors:
+#     Santiago Dueñas <sduenas@bitergia.com>
+#
+
+import click
+
+from sgqlc.operation import Operation
+
+from ..client import (SortingHatClientError,
```
mafesan: I get a `flake8` warning here (unused import).

requirements.txt
```diff
@@ -1,3 +1,4 @@
+click==7.0
```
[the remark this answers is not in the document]
sduenas: Added. Please check it.

sortinghat/cli/cmds/orgs.py
sduenas: Fixed. Please check it.

PR #263, opened 2020-03-27
[client:add] Add command to add identities

### LINKED ISSUE #256: Redefine add command
SortingHat is being redefined as a web service, so it's required to update the command line client.
On this task we will focus on updating the `add` command.

This command adds identities to the registry. Take into account the original command supports merging of matching identities. This feature is not yet implemented.

sduenas:
This command allows to create new identities in the registry.

The main differences between this and the original command is that it doesn't support matching nor merging of identities because this is not already supported.

This PR fixes #256

### DIFF
--- a/sortinghat/cli/cmds/add.py
+++ b/sortinghat/cli/cmds/add.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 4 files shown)


sortinghat/cli/cmds/add.py
```diff
@@ -0,0 +1,93 @@
...
+
+    This command adds a new identity to the registry. By default,
+    a new unique identity will also be created and linked to the
+    new identity.
+
+    The registry considers that two identities are distinct when
+    any value of the tuple (<source>, <email>, <name>, <username>)
+    is different. For example, the identities:
+
+    \b
```
mafesan: Just a comment about this character, I understand this is used for adjusting how `click` is showing this text in the command's help, right? Thanks!

sduenas: Yeah. That's right. Otherwise the paragraph below `\b` won't keep the format.

PR #264, opened 2020-03-27
[client:rm] Add command to remove identities

### LINKED ISSUE #257: Redefine rm command
The goal of this task is to redefine the rm command using the GraphQL API. The rm command removes identities and all other information related.

sduenas:
The command 'rm' allows to remove unique identities or identities from the registry.

The differences with the original command are that the flag `identity ` has been removed. When the `uuid` references to a unique identity and to an identity it will remove both. The server doesn't allow to remove only the identity.

Fixes #257

### DIFF
--- a/sortinghat/cli/templates/rm.tmpl
+++ b/sortinghat/cli/templates/rm.tmpl
@@ -1,0 +1,5 @@
+{% if is_id_removed %}
+Identity {{ uuid }} removed
+{%- else %}
+Unique identity {{ uuid }} removed
+{%- endif %}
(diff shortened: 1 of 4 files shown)


sortinghat/cli/templates/rm.tmpl
```diff
@@ -0,0 +1,5 @@
+{% if is_id_removed %}
+Identity {{ uuid }} removed
+{%- else %}
```
mafesan: The character `-` after the `%` is not present in the other templates, is there a reason for that?

sortinghat/cli/templates/rm.tmpl
```diff
@@ -0,0 +1,5 @@
...
+Unique identity {{ uuid }} removed
+{%- endif %}
```
mafesan: Same as above.

sortinghat/cli/templates/rm.tmpl
sduenas: This is a special character. When it's not set, a blank line is printed instead of the `else` control tag.

Issue #265, opened 2020-03-27
Move should only work with unique identities that have 2 or more identities


sduenas:
When `from_id` is the only identity of a unique identity, move operation should not work because a unique identity will be orphan. The operation should be `merge` instead. 

There's also another option which will consists on remove the unique identity and move only the identity.

mafesan:
I have just submitted PR https://github.com/chaoss/grimoirelab-sortinghat/pull/267 fixing this issue.

sduenas:
Fixed by #267.

PR #267, opened 2020-03-30
[api] Fix move when from identity matches its unique identity

mafesan:
This PR aims so solve issue #265. 

Before this change, two problematic cases were found:
* When `from_id` was the only identity of a UniqueIdentity, move operation made the unique identity got orphan.
* When `from_id` matched the `uuid` of its parent UniqueIndentity, the sub-identity was moved to a different UniqueIdentity (the one in `to_uuid`).

This fix prevents these cases to happen checking if `from_id` matches the `uuid` of its parent UniqueIdentity.

There was some existing tests which have been corrected after this change and a new test for this fix both for API and Schema.

### DIFF
--- a/sortinghat/core/api.py
+++ b/sortinghat/core/api.py
@@ -304,6 +304,9 @@ def move_identity(ctx, from_id, to_uuid):
     unique identity does not exist, a new unique identity will be
     created and the identity will be moved to it.
 
+    When `from_id` exists as a unique identity too, the function raises
+    an `InvalidValueError`, as this identity cannot be moved.
+
     The function raises a `NotFoundError` exception when either `from_id`
     or `to_uuid` do not exist in the registry and both identifiers are
     not the same.
@@ -335,6 +338,10 @@ def move_identity(ctx, from_id, to_uuid):
     trxl = TransactionsLog.open('move_identity', ctx)
 
     identity = find_identity(from_id)
+
+    if identity.id == identity.uidentity.uuid:
+        msg = "'from_id' is a unique identity and it cannot be moved; use 'merge' instead"
+        raise InvalidValueError(msg=msg)
 
     try:
         to_uid = find_unique_identity(to_uuid)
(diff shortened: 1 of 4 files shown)


sortinghat/core/api.py
```diff
@@ -336,6 +336,10 @@ def move_identity(ctx, from_id, to_uuid):
 
     identity = find_identity(from_id)
 
+    if identity.id == identity.uidentity.uuid:
+        msg = "'from_id' cannot match its UniqueIndentity. Try with 'merge'"
```
sduenas: Something like `'from_id' is a unique identity and it cannot be moved; use 'merge' instead` might work better. What do you think?

tests/test_schema.py
```diff
@@ -3023,13 +3047,34 @@ def test_create_new_unique_identity(self):
...
         self.assertEqual(identity_db.id, '03877f31261a6d1a1b3971d240e628259364b8ac')
         self.assertEqual(identity_db.name, None)
         self.assertEqual(identity_db.email, 'jdoe@example.com')
 
+        identity_db = identities_db[1]
+        self.assertEqual(identity_db.id, '0880dc4e621877e8520cef1747d139dd4f9f110e')
+        self.assertEqual(identity_db.name, 'J. Smith')
+        self.assertEqual(identity_db.email, 'jsmith@example.org')
+
+    def test_from_id_match_uuid(self):
```
sduenas: You can change the name of this tests to `test_from_id_is_unique_identity` or similar.

sortinghat/core/api.py
sduenas: The documentation of `move_identity` and `mv` command should be updated too.

mafesan: Done!

tests/test_schema.py
mafesan: Done!

mafesan:
Hi again @sduenas, I have addressed your comments. Thanks!

Issue #269, opened 2020-03-30
Enroll doesn't allow to create new enrollments when the default one exists


sduenas:
When a default enrolment exists for a given unique identity and organization (e.g `01234,Example, 1900-01-01, 2100-01-01`) the function `enroll` doesn't allow to create other enrollments within the same range.

This is an error because these values are considered like the min and max limits and should be replaced when a more specific date is set.

sduenas:
Fixed by #278.

PR #278, opened 2020-04-01
Allow to create new enrollments when having default dates

mafesan:
Before these changes, when a default enrollment existed for a given unique identity and organization (e.g 01234, Example, 1900-01-01, 2100-01-01) the function enroll didn't allow to
create other enrollments within the same range with more specific dates (e.g. 2004-01-01, 2006-01-01), raising an `DuplicateRangeError`.

Now, these default start and end dates are replaced when a more specific date is set if the flag `force` is active.

Note: This PR aims to fix https://github.com/chaoss/grimoirelab-sortinghat/issues/269.


sortinghat/core/api.py
```diff
@@ -607,7 +607,7 @@ def delete_domain(ctx, domain_name):
 
 
 @django.db.transaction.atomic
-def enroll(ctx, uuid, organization, from_date=None, to_date=None):
+def enroll(ctx, uuid, organization, from_date=None, to_date=None, ignore_default=False):
```
sduenas, line 610: I think we should rename `ignore_default` to `force` or something like that to overwrite default ranges. It's shorter and easier to remember - specially for the command line arguments.

sortinghat/core/errors.py
```diff
@@ -90,3 +91,10 @@ class LockedIdentityError(BaseError):
 
     code = CODE_LOCKED_IDENTITY_ERROR
     message = "UniqueIdentity %(uuid)s is locked"
+
+
+class EnrollmentRangeError(BaseError):
+    """Exception raised when setting an enrollment with a wrong
+     or conflicting date range"""
+    code = CODE_ENROLLMENT_RANGE_ERROR
```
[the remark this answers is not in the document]
mafesan: My bad, I thought this blank line was not needed when the docstring doesn't fit in one line.

sortinghat/core/errors.py
```diff
@@ -90,3 +91,10 @@ class LockedIdentityError(BaseError):
 
     code = CODE_LOCKED_IDENTITY_ERROR
     message = "UniqueIdentity %(uuid)s is locked"
+
+
+class EnrollmentRangeError(BaseError):
```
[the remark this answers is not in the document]
mafesan: I have renamed it to `DuplicateRangeError`, but I'm not totally convinced about this name. Other options could be `ExistingRangeError` or `OverlappedRangeError`. WDYT?

sortinghat/core/api.py
```diff
@@ -625,6 +625,9 @@ def enroll(ctx, uuid, organization, from_date=None, to_date=None):
 
     If the given period for that enrollment is enclosed by one already
     stored, the function will raise an `EnrollmentRangeError` exception.
+    If the option `ignore_default` is set to `True`, and the existing
+    enrollment is not composed by any of the default
+    start or end date.
```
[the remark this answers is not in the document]
mafesan: Oops! My bad :sweat_smile:

sortinghat/core/api.py
mafesan: Ok, I have renamed it also in the mutation.

sortinghat/core/errors.py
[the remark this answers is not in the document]
sduenas: `OverlappedRangeError` sounds good to me

mafesan:
@sduenas I have just submitted the changes after your comments. Thanks!

sortinghat/core/api.py
```diff
@@ -639,7 +642,7 @@ def enroll(ctx, uuid, organization, from_date=None, to_date=None):
     :raises InvalidValueError: raised in three cases, when either identity or
         organization are None or empty strings; when "from_date" < 1900-01-01 or
         "to_date" > 2100-01-01; when "from_date > to_date".
-    :raises AlreadyExistsError: raised when the given period for that enrollment
+    :raises EnrollmentRangeError: raised when the given period for that enrollment
```
sduenas, line 652: This should be `DuplicateRangeError`.

mafesan: Oops! Fixed.

sortinghat/cli/cmds/enroll.py
```diff
@@ -61,6 +64,10 @@ def enroll(ctx, uuid, organization, from_date, to_date, **extra):
 
     If the given period for that enrollment is enclosed by
     one already stored, the command will return an error.
+    This might happen when default values are already set
+    an enrollment. In the case, instead of removing an
```
mafesan, line 68: I think this should be: (...)values are already set `for`/`in` an enrollment. 
WDYT?

PR #279, opened 2020-04-01
[client:split] Add command to unmerge identities

### LINKED ISSUE #274: Add command to split unique identities
The `muggle` branch incorporates a new API method called `unmergeIdentites`. This command does the opposite `mergeIdentities` does. It splits a unique identity into different unique identities, as many as the number of identities the original had.

The goal would be to create a new command that supports this operation.

sduenas:
The new command 'split' unmerges a set of identities.

Fixes #274 .

### DIFF
--- a/sortinghat/cli/cmds/split.py
+++ b/sortinghat/cli/cmds/split.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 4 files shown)


sortinghat/cli/cmds/split.py
```diff
@@ -0,0 +1,73 @@
...
+from ..client import SortingHatSchema
+from ..utils import (connect,
+                     display,
+                     sh_client_cmd_options,
+                     sh_client)
+
+
+@click.command()
+@sh_client_cmd_options
+@click.argument('uuid', nargs=-1, required=True)
```
mafesan, line 35: I think it would be more clear if this argument is displayed as `[UUID]...` instead of `UUID...`. For instance, this is how the list of `FROM_UUID` is shown in the `merge` command:

```
sortinghat merge [OPTIONS] TO_UUID [FROM_UUID]...
```

WDYT?

sduenas: I cannot control this behaviour. `click` library does. In any case, the way that is displayed in `split` is the right one because it's required and it needs a list of arguments. `[]` does not mean list of arguments, it means optional.

I've realized when I was doing it that in `merge` the behaviour is wrong, because we always need a FROM_UUID to run `merge`. I'll fix it in a different commit.

mafesan: Oh, I see. Thanks for the clarification!

PR #280, opened 2020-04-02
[client:show] Add command to display info about unique identities

### LINKED ISSUE #276: Redefine show command
The orginal `show` command shows the information about all the unique identities stored in the registry or about a particular one defined by the user.

The goal of this task is to port that command to the new client which uses GraphQL queries.

sduenas:
The command 'show' prints to the standard output information about the unique identities on the
registry.

The main difference with the original command is that `term` argument is not available yet because the server does not support to filter by it.

Fixes #276

### DIFF
--- a/sortinghat/cli/cmds/show.py
+++ b/sortinghat/cli/cmds/show.py
@@ -1,0 +1,100 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014-2020 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


sduenas:
> @sduenas I have reviewed and testes these changes. Everything worked as expected and it LGTM!
> 
> Just a small comment: when you launch this `show` command with a non-existing `uuid`, it doesn't show any output. What do you think about a message like "uuid `<uuid`> not found in the registry", or something similar? (Like the output produced by other commands when a unique identity is not found). Thanks!

Yeah, I was thinking the same when I was implementing it but, for example, in the case of of organizations, we don't show anything. Do you think we can merge this PR and think about what kind of output do we want for all of these cases?

mafesan:
> Yeah, I was thinking the same when I was implementing it but, for example, in the case of of organizations, we don't show anything. Do you think we can merge this PR and think about what kind of output do we want for all of these cases?

@sduenas True, I didn't remember the case of `organizations`. And sure, I think we can improve this once we have all the basic cases. Let's merge this PR, thank you!

PR #287, opened 2020-04-15
Add countries support

### LINKED ISSUE #285: Add support for countries list
In muggle branch we make the model for countries data but there are some features that are missing:

* The list of countries is not loaded when the database is created
* There's no way to access the list of available countries. In the stable branch, the command `countries` provides this list.

The goal of this task is to support countries as in the stable branch. It will be needed to import the list of countries when the database is created or when the migration is applied (probably as fixtures?). Also, we will need to provide a GraphQL query and a client command to get the list of countries.

sduenas:
This PR adds three functionalities. First, it adds the list of countries that we have in the master branch. This list has been added as a fixture. Therefore, when the database is created, the list of countries can be imported using the command:

```
$ manage.py loaddata --settings <filename> countries.json
```

Second, it adds the GraphQL query `countries` to fetch the list of countries stored on the system.

Finally, the command `countries` has been also added to the client to display this list. This command allows to filter countries using the options `--code` and `--term`.

Fixes #285

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-click==7.0
+click==7.1.1
 Django==2.1
 django-mysql>=3.2.0
 django-graphql-jwt>=0.3.0
--- a/sortinghat/cli/client/schema.py
+++ b/sortinghat/cli/client/schema.py
@@ -65,6 +65,13 @@ String = sgqlc.types.String
 # Input Objects
 ########################################################################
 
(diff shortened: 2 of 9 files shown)


mafesan:
Just a brief comment about the description of this PR for future reference: The command to load the JSON data is:

```
$ python manage.py loaddata --settings <filename> countries.jsononfig.settings.devel
```

PR #294, opened 2020-05-04
Set up UI dev infrastructure

sduenas:
The purpose of this PR is to create and configure the infrastructure necessary to develop the SortingHat UI. The UI will be developed using Vue + Vuetify + Storybook. This PR provides it and adds some components to test that everything is correctly configured.

To test it you will [need to install](https://classic.yarnpkg.com/en/docs/install/#debian-stable) `yarn` 

Once you have it, install the UI dependencies with the command `yarn install` (inside the `ui` directory).

After that, run `yarn serve` and `yarn storybook`.


sduenas:
> It seems that the configuration of the ESLinter is needed (a file like `.eslintrc.js`) but is not included in the PR. I fixed the error initializing the linter directly calling the package (using `./node_modules/.bin/eslint --init` and following the steps) and this command created me the `.eslintrc.js` but I think that there should be one in the repo with the configuration.

Sorry, I forgot to include it. I've just added it and rebased the PR.

PR #295, opened 2020-05-13
[misc/grimoirelab2sh] Add new argument to avoid periods validation

zhquan:
Enrollment periods will not be valuated when
`enrollment_periods_validation` is set to `False`

Test added accordingly.


sortinghat/parsing/grimoirelab.py
```diff
@@ -50,7 +51,8 @@ class GrimoireLabParser(object):
     The keys of this object are the UUID of the unique identities.
     Each unique identity object stores a list of identities and
     enrollments. Email addresses will not be validated when `email_validation`
-    is set to `False`.
+    is set to `False`. Enrollment periods will not be valuated when
```
valeriocos, line 54: valuated -> validated

Issue #292, opened 2020-04-28
Automatic affiliation support


sduenas:
Muggle branch does not have support to enroll individuals to organizations automatically.

The goal of this task is to produce a method which will affiliate all or a set of individuals to their organizations. We can still use the same email algorithm we have been using in the master branch.

https://github.com/chaoss/grimoirelab-sortinghat/blob/master/sortinghat/cmd/affiliate.py#L71

sduenas:
As the affiliation task might take some seconds or minutes to perform, the server should run this task in the background. Also we need a recommendation module that will give the possible organizations where an individual should be enrolled. This module would help us later when we implement the matching identities system.

sduenas:
I think we are going to use `django-rq` as a task scheduler. There are many other options like `channels`, `django-q`, `celery`, `django-background-tasks` but after some evaluation, `rq` looks like the right choice. We have experience with it in Arthur, so the integration should be fast.

sduenas:
Closed by #296.

PR #296, opened 2020-05-17
Affiliation job

sduenas:
This PR addresses the request in #292 adding several features to the branch. I was tempted to split them into different PRs but I don't think it makes sense because they need to work as a whole.

* **Recommendations system engine**, which allows to generate recommendations based on the data stored in the registry. The only recommendation type available now is `affiliations`.
* **Affiliation recommender**, which  generates a list of which organizations can be enrolled to which individuals.
* **Job manager**, to schedule and execute long-running jobs, such as recommendations or affiliation, in the asynchronously.
* **Affiliation job**, which runs the affiliation process based in a series of recommendations
* **Mutation to affiliate**, which allows to schedule an affiliation job
* **Mutation to get the status of the jobs**, which giving a job id returns the information about that job.

This PR adds a new dependency: `django_rq`. This  [library](https://github.com/rq/django-rq) provides the job manager to schedule and run jobs. You will need to add this configuration in the Django config file (see the `testing.py` file for an example):
```
INSTALLED_APPS = [
    'django_rq',
    'graphene_django',
    'sortinghat.core',
    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
]


RQ_QUEUES = {
    'default': {
        'HOST': 'localhost',
        'PORT': 6379,
        'DB': 0
    }
}
```
RQ uses Redis, so you will need to have a Redis server running or use `fakeredis` to pretend you are running it. If that's the case,  add this to your configuration (**before** the declaration of `RQ_QUEUES`).

```import django_rq.queues

from fakeredis import FakeRedis, FakeStrictRedis

class FakeRedisConn:
    """Singleton FakeRedis connection."""

    def __init__(self):
        self.conn = None

    def __call__(self, _, strict):
        if not self.conn:
            self.conn = FakeStrictRedis() if strict else FakeRedis()
        return self.conn

django_rq.queues.get_redis_connection = FakeRedisConn()

RQ_QUEUES = {
    'default': {
        'HOST': 'localhost',
        'PORT': 6379,
        'ASYNC': False,
        'DB': 0
    }
}
```

If you prefer to run it as in production, you will need to run some RQ workers to perform the jobs. You can run them with the command:
```
./manage.py rqworker --settings=<config>
```

To test the PR, add some organizations and individuals; call the mutation `affiliation` and check the results calling `job` query.

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -53,7 +53,7 @@ from .api import (add_identity,
                   withdraw)
 from .context import SortingHatContext
 from .decorators import check_auth
-from .jobs import affiliate
+from .jobs import affiliate, find_job
 from .models import (Organization,
                      Domain,
                      Country,
@@ -134,6 +134,23 @@ class ProfileType(DjangoObjectType):
 class EnrollmentType(DjangoObjectType):
     class Meta:
         model = Enrollment
+
+
+class AffiliationResultType(graphene.ObjectType):
+    uuid = graphene.String()
+    organizations = graphene.List(graphene.String)
+
+
+class JobResultType(graphene.Union):
+    class Meta:
+        types = (AffiliationResultType,)
+
+
+class JobType(graphene.ObjectType):
+    job_id = graphene.String()
+    status = graphene.String()
+    result = graphene.List(JobResultType)
+    errors = graphene.List(graphene.String)
 
 
 class ProfileInputType(graphene.InputObjectType):
@@ -581,6 +598,10 @@ class SortingHatQuery:
         page=graphene.Int(),
         filters=OperationFilterType(required=False),
     )
+    job = graphene.Field(
+        JobType,
+        job_id=graphene.String()
+    )
 
     @check_auth
     def resolve_countries(self, info, filters=None,
@@ -626,6 +647,28 @@ class SortingHatQuery:
         return IdentityPaginatedType.create_paginated_result(query,
                                                              page,
                                                              page_size=page_size)
+
+    @check_auth
+    def resolve_job(self, info, job_id):
+        job = find_job(job_id)
+
+        status = job.get_status()
+        result = job.result
+
+        if job.result:
+            errors = result['errors']
+            affiliated = [
+                AffiliationResultType(uuid=uuid, organizations=orgs)
+                for uuid, orgs in result['results'].items()
+            ]
+        else:
+            errors = None
+            affiliated = None
+
+        return JobType(job_id=job_id,
+                       status=status,
+                       result=affiliated,
+                       errors=errors)
 
     @check_auth
     def resolve_transactions(self, info, filters=None,
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -30,13 +30,13 @@ from django.core.paginator import Paginator
 
 from django_mysql.models import JSONField
 
+from django_rq import enqueue
+
 from graphene.types.generic import GenericScalar
 
 from graphene_django.converter import convert_django_field
 from graphene_django.types import DjangoObjectType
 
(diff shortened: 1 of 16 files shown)


tests/test_schema.py
```diff
@@ -4799,3 +4966,174 @@ def test_authentication(self):
...
+        """Check if all the individuals stored in the registry are affiliated"""
+
+        mock_job_id_gen.return_value = "1234-5678-90AB-CDEF"
+
+        client = graphene.test.Client(schema)
+
+        executed = client.execute(self.SH_AFFILIATE,
+                                  context_value=self.context_value)
+
+        # Check results, identities were unmerged
```
mafesan, line 5052: Update this comment to `Check results, identities were affiliated`.

sortinghat/core/schema.py
```diff
@@ -133,6 +136,23 @@ class Meta:
         model = Enrollment
 
 
+class AffiliationResultType(graphene.ObjectType):
+    uuid = graphene.String()
+    organizations = graphene.List(graphene.String)
```
mafesan, line 141: I think `organizations` should be defined as a list of `OrganizationType`. Is there a specific reason for setting it as a list of strings?

tests/test_jobs.py
```diff
@@ -0,0 +1,293 @@
...
+                         uuid=self.jroe.uuid)
+        api.add_identity(ctx,
+                         source='unknown',
+                         email='jroe@bitergia.com',
+                         uuid=self.jroe.uuid)
+
+    def test_affiliate(self):
+        """Check if all the individuals stored in the registry are affiliated"""
+
+        ctx = SortingHatContext(self.user)
```
mafesan, line 131: I have a question here: It is necessary to create a new context before performing each request? Could the context defined in the `setUp` be used for all the requests?

sortinghat/core/schema.py
```diff
@@ -605,6 +648,28 @@ def resolve_individuals(self, info, filters=None,
...
+            errors = result['errors']
+            affiliated = [
+                AffiliationResultType(uuid=uuid, organizations=orgs)
+                for uuid, orgs in result['results'].items()
+            ]
+        else:
+            errors = None
+            affiliated = None
+
+        return JobType(job_id=job_id,
```
mafesan, line 668: Although paginating jobs does not make sense here (as we are asking for one job), would it be possible to paginate the `AffilitationResultType` results somehow?

mafesan:
Just for the record: I have realized my testing settings were partially wrong, as the definition of the `Singleton FakeRedis connection` has to be placed on the bottom of the script (I placed it on top of it). That's why some tests were working and others were failing.

tests/test_jobs.py
sduenas: Probably it's not necessary but I wanted to be user I was using a new context each time.

sortinghat/core/schema.py
sduenas: It would be possible but we have to think how. The results are stored in Redis all in the same structure which means we will have to do the pagination once the results are gathered by the django view. I think we can leave it this way not and think about possible solutions, if needed, for the future.

sortinghat/core/schema.py
sduenas: The results generated by the job return the name of the organizations an individual was affiliated. As the results are stored in a Redis queue, strings work better that objects because with complex objects we would have to define methods to marshall and unmarshall when they are read/written.

What we can do is to convert the name to an `OrganiationType` in the graphene resolver but I not sure how to do it now. Let's keep this as is now and let's think about it to improve this feature.

sduenas:
Please check the changes in the PR.

Issue #303, opened 2020-06-10
[ui] Support pagination in individuals query


sduenas:
The current query only fetches the first page of individuals. This query should be able to request different pages and specify the number of items to retrieve on each query:

Here is the implementation of the pagination type in GraphQL.
https://github.com/chaoss/grimoirelab-sortinghat/blob/muggle/sortinghat/core/schema.py#L77

dlumbrer:
After thinking about the issue, I submitted a PR (#307) that adds the functionality of the pagination to the query (test included).
I didn't touch anything related to the UI because I think that that functionality falls on #306. 

Please @sduenas , check my comment. If you agree, I will work on the other issue and adapt the components to use the pagination functionality (after mocking up the design as we talked about).

sduenas:
Fixed by #307

PR #307, opened 2020-06-30
[ui] Support pagination arguments in getIndividuals query

dlumbrer:
This PR allows the getIndividuals query to use the pagination functionality, adding the next arguments:

- page: integer in order to ask the page (by default, 1)
- pageSize: integer in order to define the size of the query (by default, 10) 

Moreover, the query returns all the data related to the pagination (pageInfo).

This PR is related to #303

### DIFF
--- a/ui/tests/unit/queries.spec.js
+++ b/ui/tests/unit/queries.spec.js
@@ -24,6 +24,16 @@ describe("App", () => {
             }
           ],
           __typename: "IdentityPaginatedType"
+        },
+        pageInfo: {
+          page: 1,
+          pageSize: 10,
+          numPages: 1,
+          hasNext: false,
+          hasPrev: false,
+          startIndex: 1,
+          endIndex: 1,
+          totalResults: 1
         }
       }
     };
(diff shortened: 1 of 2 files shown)


ui/tests/unit/queries.spec.js
```diff
@@ -24,6 +24,16 @@ describe("App", () => {
             }
           ],
           __typename: "IdentityPaginatedType"
+        },
```
sduenas, line 27: I think we should include a test to evaluate the parameters we are passing to the method are correctly set.

dlumbrer:
I updated the PR @sduenas , take into account that these tests may change when we change the pagination in the UI with the components, so these tests are the first approach.

ui/tests/unit/queries.spec.js
```diff
@@ -2,31 +2,43 @@ import { shallowMount } from "@vue/test-utils";
 import Vue from "vue";
 import Vuetify from "vuetify";
 import App from "@/App";
-import { getIndividuals } from "@/apollo/queries";
+//import { getIndividuals } from "@/apollo/queries";
+import * as Queries from "@/apollo/queries";
```
sduenas, line 6: Why did you decide to leave this comment?

dlumbrer: Sorry, I forgot to remove it!

dlumbrer:
> The patch looks good. Can you respond to the other comment before we merge this?

I updated the PR, I forgot to remove the comment, sorry about it.

Issue #302, opened 2020-06-08
Matching algorithm for muggle


sduenas:
In the first version of sortinghat, this was done with the `match` algorithms, that compared a field of set of fields between the identities to find cases where those values were the same. If they were the same, it considered the unique identities were the same too. In the case of `unify` the matching was done using `pandas` to improve the speed of the process.

In `muggle` branch we need to define how to do this process. I think we should follow the the same approach using `pandas` but we need a way to compare a subset with another bigger subset. The original algorithm only allows to compare a set with the same set.

Also, the algorithm should be a comparing field algorithm where the input will be a list of fields to compare.

mafesan:
Adding some context about the existing matching algorithm, which could be summarized in these main steps:
* Convert the input set to `pandas` *dataframes* filtering by the specified matching criteria. E.g. If the matching criteria is by using the `email` field, all the input values without an email will be dropped.
* Form groups using the different input identities.
* Calculate the matches iterating over the resulting groups.
* The result is composed by groups of matching identities found in the whole input set (all identities are compared among them). Identities with no matches are returned as one-element groups.

Before providing a valid solution, these are the initial ideas that were explored:

* Instead of calculating the matching for each criteria, the idea was to provide all the criteria at once as merge method accepts a list of columns for the on parameter.

  * Result: Using a list in `on` parameter handles the columns with an AND operation and this behavior is not valid. In order to handle them as an OR comparison, we will have to keep a loop iterating over each criteria.

* Among the different merging methods, we were using the `left` method. The idea was to test with `inner` method, as it is defined as the intersection between both sets.

  * Result: Changing the merging method to `inner` discards those input results which don't have any matches. This behavior is valid.

Having said this, given two sets of identities (set A and set B), the output of this method would be a list of those identities from set A having a list of matching results with the identities from set B given a criteria.

Note: There is a behavior to be noted when performing a match operation: fields have to be normalized before performing the operation (maybe at the moment when these values are added to the registry at first place). For instance, these two emails won't be considered as a match (and they should): `jsmith@example.com`, `JSmith@example.com`.

mafesan:
I'm preparing a PR with a proposal for this new version of the algorithm version and it will be included in SortingHat as a recommendation, just like `affiliation`. The idea is given a list of recommendations about which identities match with a given set of individuals (or its identities) so in a further step it can be decided whether applying the changes (merging them) or not.

[...]

PR #308, opened 2020-07-10
Get matching identities recommendations from sets of individuals

mafesan:
Recommend identity matches for a list of individuals.

Note: This PR is related to issue #302 .

This job generates a list of recommendations with the matching identities which the input individuals can be merged with, based on a list of criteria composed by email addresses, name and/or usernames of the individuals.

The function behind the scenes checks if any identity from each individual matches with a given set of target individuals. First, it filters by the fields from the criteria and then it groups the matches by identity. Then, these groups are turned into sets of matching identities found among all the input identities set.

Each recommendation contains the uuid of the individual provided in the input list and a list with the matching individuals or the matching identities if the `verbose` option is activated.

When no matching is found for a given individual, an empty list will be returned for that individual.

Individuals both for `source_uuids` and `target_uuids` are defined by any of their valid keys or UUIDs. When the parameter `target_uuids` is empty, the job will take all the individuals stored in the registry, so matches will be found comparing the identities from the individuals in `source_uuids` against all the identities on the registry.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,3 +14,4 @@ jinja2==2.11.1
 rq>=1.4.0
 django-rq>=2.3.2
 fakeredis>=1.4.1
+pandas>=1.0.5
--- a/sortinghat/core/jobs.py
+++ b/sortinghat/core/jobs.py
@@ -17,6 +17,7 @@
 #
 # Authors:
 #     Santiago Dueñas <sduenas@bitergia.com>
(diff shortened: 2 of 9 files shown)


mafesan:
I have changed one behavior: when no matches are found for a given individual, the method now returns an empty list, just like the affiliation method (before, when no matches were found for a given individual, the individual was not included in the results).

mafesan:
Thank you for the review @sduenas ! I have just submitted the changes I made after your comments.

Issue #310, opened 2020-07-20
Support automatic merge of matching identities


mafesan:
Muggle branch does not have support to merge matching individuals automatically.

After having the matching recommendations (see issue #308), the goal of this task is to produce a method which will merge all or a set of individuals to the individuals from the matching results.

The approach should be similar to the work done for the automatic affiliation (see issue #292).

sduenas:
Closed by #312

PR #312, opened 2020-07-29
Unify matching identities automatically

mafesan:
## Unify a set of individuals by merging them using matching recommendations.

Note: this PR is related to https://github.com/chaoss/grimoirelab-sortinghat/issues/310.

This function automates the identities unify process obtaining a list of recommendations where matching individuals can be merged. After that, matching individuals are merged.
The corresponding job returns a list with the individuals which have been merged and the errors generated during this process.

The `unify` mutation returns a job which returns a list of identities which have been merged with other matching identities according to the specified criteria (a combination of email, name and username).

If the target list of individuals is empty, the job will be applied to all the individuals in the registry.

Example for the new mutation:
```
mutation {
  unify(sourceUuids: ["<uuid1>",
                      "<uuid2>"],
        criteria: ["email", "name", "username"]){
    jobId
  }
}
```

Example for the query asking for the job result:
```
{
  job(jobId: "<job_id>"){
    jobId
    status
    jobType
    result {
      __typename
      ... on UnifyResultType {
          merged
      }
    }
    errors
  }
}
```

There was a problem when groups contain identities with different identifiers. This causes they can't be grouped at the first place, so when the merge method is called, it raised an `InvalidValueError` when source identity was already part of the destination. To avoid this behavior, a new error was created: `EqualIndividualError`, which is raised only in this case. The `unify` method ignores this case, as in this scenario this could not be considered as an error, but a situation where there are no changes to apply.

Apart from these main changes, the matches recommendation function has been modified to remove key UUID from matching results. Before this change, an UUID used as a key could be included in the set of matching results. Now, it is filtered out before returning the results.


tests/test_jobs.py
```diff
@@ -523,8 +524,7 @@ def test_recommend_matches(self):
         # Test
         expected = {
             'results': {
-                self.john_smith.uuid: sorted([self.john_smith.uuid,
```
mafesan, line 526: This is an example of what I've explained in the text of [this commit](https://github.com/chaoss/grimoirelab-sortinghat/pull/312/commits/7cb6238c7643189cf125b81beab91eaa4efd652e).

tests/test_jobs.py
```diff
@@ -724,3 +720,346 @@ def test_transactions(self):
...
+        criteria = ['email', 'name']
+
+        job = unify.delay(ctx,
+                          source_uuids,
+                          target_uuids,
+                          criteria)
+        result = job.result
+
+        self.assertDictEqual(result, expected)
+
```
mafesan, line 1033: I was not sure about what to to with this test. It was meant to test what happens when an input individual is not found on the registry. However, when this happens it causes an error in the recommendation system and that is not captured here. If any errors were returned here, they would have been related with the merging process... WDYT @sduenas ?

sduenas: Maybe the method should return the errors produced during the recomendation? Does it make sense to you?

mafesan:
Hi @sduenas, I have addressed most of your comments and I have modified some parts of this PR too.

As suggested, it had no sense for the `_find_matches` method to return different results depending on the option `grouped`. I removed this and this grouping is now performed within the job method, after obtaining the results from the matching recommendations.

Both this and the former complex solution are caused mainly because there was a problem when groups contained identities with different identifiers (individuals can be referenced using any of their UUIDs). When this happens, they can't be grouped at the first place so when the merge method is called, the merge method raised an `InvalidValueError` when source identity was already part of the destination. 

To avoid this behavior, a new error was created: `EqualIndividualError`, which is raised only in the case that I mentioned. The unify method ignores this case, as in this scenario this could not be considered as an error, but a situation where there are no changes to apply.

sortinghat/core/recommendations/matching.py
```diff
@@ -29,7 +29,7 @@
 from ..errors import NotFoundError
 
 
-def recommend_matches(source_uuids, target_uuids, criteria, verbose=False):
+def recommend_matches(source_uuids, target_uuids, criteria, verbose=False, groupped=False):
```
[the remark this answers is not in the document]
mafesan: This is no longer valid  (see comment https://github.com/chaoss/grimoirelab-sortinghat/pull/312#issuecomment-669106301).

sortinghat/core/recommendations/matching.py
```diff
@@ -172,6 +205,33 @@ def _calculate_matches_groups(grouped_uids, verbose=False):
 
         return matches
 
+    def _calculate_unique_matches(grouped_uids):
```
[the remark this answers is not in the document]
mafesan: This is no longer valid  (see comment https://github.com/chaoss/grimoirelab-sortinghat/pull/312#issuecomment-669106301).

tests/test_jobs.py
mafesan: I agree, but I couldn't find a simple way to include these errors in the result without modifying the recommendation engine. Do you think we should modify the engine to support it?

sortinghat/core/api.py
```diff
@@ -868,7 +869,7 @@ def _find_individuals(from_uuids, to_uuid):
             if from_uuid == '':
                 raise InvalidValueError(msg="'from_uuid' cannot be an empty string")
             if from_uuid == to_uuid:
-                raise InvalidValueError(msg="'from_uuid' and 'to_uuid' cannot be equal")
+                raise EqualIndividualError(msg="'from_uuid' and 'to_uuid' cannot be equal")
```
sduenas, line 872: I'm not sure if this is the best error message because the function receives `from_uuids` which is a list. I think it should be better something like `to_uuid cannot be part of `from_uuids`. I also think it would be useful to have the hash `uuid` in the message.

sortinghat/core/jobs.py
```diff
@@ -213,6 +213,126 @@ def affiliate(ctx, uuids=None):
...
+        :param recs: recommendations of matching identities
+
+        :returns: a list including unique groups of matches
+        """
+        groups = []
+        for group_key in recs:
+            g_uuids = recs[group_key]
+            if group_key not in g_uuids:
+                g_uuids.append(group_key)
+            g_uuids = sorted(g_uuids)
```
sduenas, line 258: What's the point of sorting the uuids here? This can slow down the function.

sortinghat/core/jobs.py
```diff
@@ -213,6 +213,126 @@ def affiliate(ctx, uuids=None):
...
+            if (len(g_uuids) > 1) and (g_uuids not in groups):
```
sduenas, line 259: Why do we need to check if the `g_uuids` are in `groups`?

sortinghat/core/jobs.py
```diff
@@ -213,6 +213,126 @@ def affiliate(ctx, uuids=None):
...
+
+    :param job_ctx: job context
+    :param source_indv: valid individual identifier where
+        the rest of individuals will be merged to
+    :param target_indvs: list of identifiers of the individuals
+        who will be merged with the source individual
+
+    :returns: tuple with the uuids from the individuals who
+        were merged and list of errors found during the process
+    """
```
sduenas, line 318: This comment doesn't match with what the function is returning. It only returns one uuid.

sortinghat/core/jobs.py
mafesan: This is because the same matching group could exist for different `group_key`, but as the result does not include the uuid from the key, it is necessary to add the key with the rest of the result group and sort it to make sure it has not been included before.

For instance, you could have the groups:
{A: [B, C], C: [B, A]}.
First, when the loop iterates over the keys, it includes that key in the list of results ([B, C, A] in the first case, [B, A, C] in the other case). To prevent adding this set duplicate, it is sorted and then checked if the set is already in the list (both would be sorted to [A, B, C]).

This comment also answers to why is necessary to check if `g_uuids` are already in `groups` or not (https://github.com/chaoss/grimoirelab-sortinghat/pull/312#discussion_r465647004).

I have updated this piece of code with Pandas methods, so at least the sorting and the `unique` operation are more efficient.

sortinghat/core/jobs.py
mafesan: Please check my answer to this comment: https://github.com/chaoss/grimoirelab-sortinghat/pull/312#discussion_r466906138.

mafesan:
> In overall it looks ok but please check my comments.
> I'd also like to have a benchmark of the this method and the previous one to see if we have improved the speed of this job. Is it possible?

Sure! Luckily I saved a copy of the former version so I could perform the tests (Note: the benchmarks are obtained after replacing the piece of code forming the unique groups using Pandas methods).

I performed the profiling for both old and new versions and in two different databases. One of the databases has around 1100 identities and the other one around 44000. The test consisted on launching the `unify` method with 10 input identities against the whole registry. These are the results:

| Version | Total number of uids | Total time (secs)|
| ---------- | --------- | -------- |
| old | ~1100 | 12.445 |
| new | ~1100 | 6.666 |
| old | ~44000 | 4372.074 |
| new | ~44000 | 3778.876 |

Whilst in the small database the difference is more significant, there is still some improvement with the larger database. Please, tell me if you need me to provide more detailed results (I can upload the `cprofile` output files too).

Thank you @sduenas !

mafesan:
Looking at the benchmarks obtained for the `unify` method, it turns out most of the time was consumed with queries executed from the `recommendations` method.  More specifically, in the case when the method receives an empty list of `target_uuids`, because the whole set of identities was retrieved in an inefficient way, as the former input was a query set with all the individuals on the registry. 

Each individual caused two queries, one for retrieving the individual object and another one for getting its identities. This was slowing down the performance heavily.

I have fixed this method so when this case happens all the identities from the registry are retrieved with a single, more efficient query. Before these last changes, the `target_uuids` were retrieved at `jobs` level and now the `recommend_matches` method is in charge of this operation.

I.e., replicating the tests for the `matching` section of the method, the results improved from ~3200 seconds to 5.186 seconds.

Issue #314, opened 2020-09-17
Warnings are raised running ui tests


sduenas:
When running the `ui` tests on `muggle` branch I get some warnings. The tests should run without raising any warning.

```
$ yarn test:unit
yarn run v1.16.0
$ vue-cli-service test:unit
 PASS  tests/unit/queries.spec.js
  ● Console

    console.error node_modules/vue/dist/vue.runtime.common.dev.js:621
      [Vue warn]: Avoid adding reactive properties to a Vue instance or its root $data at runtime - declare it upfront in the data option.
    console.error node_modules/vue/dist/vue.runtime.common.dev.js:621
      [Vue warn]: Error in created hook (Promise/async): "TypeError: Cannot read property 'data' of undefined"
      
      found in
      
      ---> <App>
             <Root>
    console.error node_modules/vue/dist/vue.runtime.common.dev.js:1884
      TypeError: Cannot read property 'data' of undefined
          at VueComponent.created (/home/sduenas/devel/grimoirelab/sortinghat/ui/src/App.vue:58:1)
          at process._tickCallback (internal/process/next_tick.js:68:7)

 PASS  tests/unit/storybook.spec.js
  ● Console

    console.error node_modules/vue/dist/vue.common.dev.js:630
      [Vue warn]: Property or method "source" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.
      
      found in
      
      ---> <Anonymous>
             <VApp>
               <Anonymous>
                 <Root>

```

sduenas:
Fixed by #315.

PR #315, opened 2020-09-18
Fix ui test warnings

evamillan:
This PR fixes the three Vue warnings that are raised when running the ui tests described on #314 .
Two of them were due to some variables not having been declared before being used or assigned a value, and one was caused by a failing query on App.vue that didn't have any error control.


sduenas (approved):
LGTM

Issue #321, opened 2020-09-22
[ui] Add visual cue to indicate when an individual is locked


sduenas:
The new version of SortingHat allows to **lock** individuals so no operations will modify them. You won't be able to delete, add new identities, enroll, etc. The only operation will be to **unlock** it.

The component should show when an individual is locked and when it's not. A lock icon on the top right corner might work. Take into account the next feature will be to be able to lock and unlock an individual from the ui.

sduenas:
Fixed by #323

PR #323, opened 2020-09-22
[ui] Add locked/unlocked icon to IndividualCard

evamillan:
This PR adds an icon to the 'IndividualCard' component
to show whether or not the profile is locked as requested in #321 .
This information can be passed through the prop 'isLocked' from the 'ProfileCard'
and 'IndividualsGrid' components.

imagen


sduenas (approved):
LGTM

Issue #322, opened 2020-09-22
[ui] Allow to lock individuals


sduenas:
The UI should allow to protect individuals from modifications. This new feature will work as follows:

1. The user can click on an icon in the top corner of the card.
2. The app will call to `lock` SortingHat method, passing the `uuid` of the individual.
3. If the operation is successful, the UI will show that the individual has been protected using the "lock icon" described in #321
4. If the operation fails, the UI will show an error message

sduenas:
Fixed by #324

PR #324, opened 2020-09-23
[ui] Allow to lock individuals

evamillan:
This PR adds the functionality of locking or unlocking
individuals by clicking on the lock icon on the `IndividualCard`
component requested in #322 .
Those mutations require the individual's `uuid` as an
argument, so it is added to `IndividualCard` as a prop and passed
from `IndividualsGrid` and `ProfileCard`.

If the mutation fails, a snackbar with the error message is
displayed on the bottom of the screen.

### DIFF
--- a/ui/src/components/IndividualCard.vue
+++ b/ui/src/components/IndividualCard.vue
@@ -19,17 +19,21 @@
 
       <v-list-item-icon>
         <v-btn text icon>
-          <v-icon small>
-            {{ isLocked ? "mdi-lock" : "mdi-lock-open-outline" }}
+          <v-icon small @click="toggleLockedStatus">
+            {{ locked ? "mdi-lock" : "mdi-lock-open-outline" }}
           </v-icon>
         </v-btn>
       </v-list-item-icon>
     </v-list-item>
-    <slot />
+    <v-snackbar v-model="snackbar.value" color="error">
+      {{ snackbar.text || "Error" }}
+    </v-snackbar>
   </v-card>
 </template>
 
 <script>
+import { lockIndividual, unlockIndividual } from "../apollo/mutations";
+
 export default {
   name: "individualcard",
   props: {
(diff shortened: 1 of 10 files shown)


sduenas (changes requested):
In overall, the PR looks ok. We need to add some tests before we merge it, though.

We need:
* Storybook tests to check if the transition from one state to another (locked - unlocked or vice-versa) works.
* Unit tests to check that the call to the server works and it applies the right state. You will need to mock the server response.

evamillan:
@sduenas I updated the PR with the changes you requested.

ui/src/components/IndividualCard.vue
```diff
@@ -69,6 +86,37 @@ export default {
       } else {
         return `mdi-${datasource}`;
       }
+    },
+    async toggleLockedStatus() {
+      if (this.$apollo) {
+        try {
+          const mutation = this.locked
+            ? unlockIndividual(this.$apollo, this.uuid)
+            : lockIndividual(this.$apollo, "def");
```
sduenas, line 95: There's an error in here. When clicking to `lock` an `Individual`, the function passes `def` instead of the `uuid`.

sduenas (changes requested):
There's an error on your code. Please fix it. The rest of the PR looks good to me.

ui/src/components/IndividualCard.vue
evamillan: Thanks for noticing, I fixed it.

sduenas: I see you have changed it. The reason why we didn't catch this before is because we're not validating the parameters we're passing to the mock functions. This validation should be done on the unit tests. Can you add some assertions? In Python there are several methods that do that, for example : https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_with . I guess there's something similar on `jest`.

evamillan: I added the assertions to the "Lock individual on click" and "Unlock individual on click" tests.

sduenas (changes requested):
I got errors running `yarn test:unit`. I think you need to update the snapshot.

evamillan:
You're right. I updated the snapshots.

sduenas (approved):
LGTM

PR #332, opened 2020-10-01
[ui] Implement IndividualEntry component

evamillan:
Adds the `IndividualEntry` component, which displays a table row with an individual's name, organization, email and sources. The information can be expanded, but the slot for the expanded content must be implemented inside the parent `<v-table>` component and not this one. I mocked the behaviour in the Storybook story, placing `IndividualEntry` inside a table and adding the expanded content there.

### DIFF
--- a/ui/src/components/IndividualEntry.stories.js
+++ b/ui/src/components/IndividualEntry.stories.js
@@ -41,13 +41,78 @@ export const Default = () => ({
       { value: 'email' },
       { value: 'sources' },
       { value: 'actions' }
-
     ],
     items: [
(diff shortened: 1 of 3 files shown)


sduenas (changes requested):
The PR looks ok but we need to have stories for three extra cases to see how individuals will be displayed. These are the cases:

- Individuals might not have organizations.
- Individuals might not have and email address set.
- Individuals might only have a first name (`Tom`) so the avatar should display only that initial.

evamillan:
I added the extra stories.

Issue #330, opened 2020-10-01
[ui] Implement Individuals table basic component


sduenas:
We finally decided how to visualize the list of individuals managed by SortingHat. It's going to be a table. Each entry will show basic information of each Individual: name, email, data sources, and affiliation. This is a mock up of the table.

![imagen](https://user-images.githubusercontent.com/833352/94785875-8c7cc480-03d0-11eb-8fc6-1a59e84bcb1f.png)

Right now we only need to implement it to show the whole list of Individuals. Search or filter buttons will be implemented later.

sduenas:
Fixed by #333.

PR #333, opened 2020-10-05
[ui] Implement Individuals table basic component

evamillan:
Adds the `IndividualsTable` component, which displays a
list of `IndividualEntry` rows that can be expanded to
show an individual's identities and affiliations as requested
on #330


sduenas (changes requested):
The PR looks ok but I think `ExpandedIndividual` should have its own story. We need to visualize how does look like and tests what will happend if an `Individual` doesn't have affiliations.

Please also rebase this PR with `muggle`.

evamillan:
I added the stories for `ExpandedIndividual`and rebased with `muggle`

PR #335, opened 2020-10-06
[ui] Add pagination to Individuals table 

evamillan:
Adds pagination to `IndividualsTable` and embeds the component into `App`. It also adds the `getPaginatedIndividuals` query to retrieve the paginated data needed for the table and tests it.

### DIFF
--- a/ui/src/components/IndividualsTable.vue
+++ b/ui/src/components/IndividualsTable.vue
@@ -1,29 +1,40 @@
 <template>
-  <v-data-table
-    hide-default-header
-    hide-default-footer
-    :headers="headers"
-    :items="individuals"
-    :expanded.sync="expandedItems"
-    item-key="id"
-  >
-    <template v-slot:item="{ item, expand, isExpanded }">
-      <individual-entry
-        :name="item.name"
-        :organization="item.organization"
-        :email="item.email"
-        :sources="item.sources"
-        :is-expanded="isExpanded"
-        @expand="expand(!isExpanded)"
-      />
-    </template>
-    <template v-slot:expanded-item="{ item }">
-      <expanded-individual
-        :enrollments="item.enrollments"
-        :identities="item.identities"
-      />
-    </template>
-  </v-data-table>
+  <v-container>
+    <v-data-table
+      hide-default-header
+      hide-default-footer
+      :headers="headers"
+      :items="individuals"
+      :expanded.sync="expandedItems"
+      item-key="id"
+      :page.sync="page"
+    >
+      <template v-slot:item="{ item, expand, isExpanded }">
+        <individual-entry
+          :name="item.name"
+          :organization="item.organization"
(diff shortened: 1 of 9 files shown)


ui/src/components/IndividualsTable.vue
```diff
@@ -35,23 +46,102 @@ export default {
   components: {
     IndividualEntry,
     ExpandedIndividual
- },
+  },
   props: {
-    individuals: {
-      type: Array,
+    query: {
```
sduenas, line 51: Can we change the name of this? It's not really a query and it's not querying, so it's a bit confusing. Maybe we can rename it to `fetchPage` or `fetch`? I'm not sure what the conventions are for naming props functions in Vue.

ui/src/components/IndividualsTable.vue
```diff
@@ -35,23 +46,102 @@ export default {
...
+        } else {
+          if (result.others) {
+            result.others.identities.push(val);
+          } else {
+            result.others = {
+              name: "Others",
+              identities: [val]
+            };
+          }
+        }
```
sduenas, line 130: Wouldn't have more sense to this only once? For example (syntax can be wrong):
```
const source = val.source.toLowerCase();
const name = source;

if (!icons.find(icon => icon === source)) {
  source = "others";
  name = "Others";
}

if (result[source]) {
  result[source].identities.push(val);
} else {
  result[source] = {
    name: source,
    identities: [val]
  };
}
```

sduenas (changes requested):
Please check my comments. I think they make sense for this context.

evamillan:
They make a lot of sense, I made the changes.

sduenas (changes requested):
The PR looks ok but I got some warnings running `yarn serve`:

```
/home/sduenas/devel/grimoirelab/sortinghat/ui/src/components/ExpandedIndividual.vue
   4:12  warning  Replace `·v-for="(source,·sourceIndex)·in·identities"·:key="source.name"·class="indented">` with `⏎······v-for="(source,·sourceIndex)·in·identities"⏎······:key="source.name"⏎······class="indented"`            prettier/prettier
   5:5   warning  Replace `····<v-list-item·v-for="(identity,·index)·in·source.identities"·:key="identity.uuid">` with `>⏎······<v-list-item⏎········v-for="(identity,·index)·in·source.identities"⏎········:key="identity.uuid"`  prettier/prettier
   6:3   warning  Insert `····>⏎`                                                                                                                                                                                                  prettier/prettier
   7:11  warning  Delete `··`                                                                                                                                                                                                      prettier/prettier
   8:1   warning  Replace `··············` with `············`                                                                                                                                                                     prettier/prettier
   9:11  warning  Delete `··`                                                                                                                                                                                                      prettier/prettier
```

sduenas (approved):
LGTM

PR #342, opened 2020-10-08
[ui] Implement basic workspace area

evamillan:
Adds the `WorkSpace` component, which displays a set of individual cards. The individuals can be saved to the work space by clicking the pin button on `IndividualEntry` or dragging and dropping the table row over the space. The work space can be cleared by clicking a button.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -34,9 +34,19 @@
     </v-app-bar>
 
     <v-content>
+      <work-space
+        :individuals="savedIndividuals"
+        @clearSpace="savedIndividuals = []"
+      />
       <v-container max-width="900">
-        <individuals-table :fetch-page="getIndividualsPage" />
+        <individuals-table
+          :fetch-page="getIndividualsPage"
+          @saveIndividual="addSavedIndividual"
+        />
       </v-container>
+      <v-snackbar v-model="snackbar">
+        Individual already in work space
+      </v-snackbar>
     </v-content>
   </v-app>
 </template>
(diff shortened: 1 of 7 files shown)


ui/src/App.vue
```diff
@@ -34,26 +34,59 @@
...
+    },
+    async addSavedIndividual(uuid) {
+      const isSaved = this.savedIndividuals.find(individual => {
+        return individual.identities[0].uuid === uuid;
+      });
+      if (isSaved) {
+        this.snackbar = true;
+        return;
+      }
+      const response = await getProfileByUuid(this.$apollo, uuid);
```
sduenas, line 85: Is this needed? Can't we reuse the information that is already on the entry? The less we visit the server, the better. Although, if it's too complicated, we can leave it this way.

ui/src/components/WorkSpace.stories.js
```diff
@@ -0,0 +1,132 @@
+import { storiesOf } from "@storybook/vue";
```
sduenas, line 1: Would be possible to add a test to simulate the drag and drop?

sduenas (changes requested):
Please check my comments. We can discuss about them.

ui/src/App.vue
evamillan: I had to make a few changes and move the logic around a bit for it to work without the request, but it's better not to visit the server a second time for the same info.

evamillan:
I updated the PR with the changes you requested.

ui/src/components/WorkSpace.stories.js
sduenas: Great. I see it working.

sduenas (approved):
LGTM

PR #345, opened 2020-10-09
[ui] Implement a OrganizationEntry component

evamillan:
Adds the `OrganizationEntry` component, which displays a table row with an organization's name and number of individuals enrolled.


sduenas (approved):
LGTM

Issue #347, opened 2020-10-13
[ui] Remove individuals table action


sduenas:
The table of individuals should allow to remove single or a group of individuals from the registry.

sduenas:
Fixed by #349.

PR #349, opened 2020-10-14
[ui] Remove individuals table action

evamillan:
Adds the `deleteIdentity` mutation, which removes an individual. The individual can be deleted from the actions menu in `IndividualEntry` or from the table actions when it's selected on `IndividualsTable`.

A dialog appears before the individual is deleted that prompts the user to confirm the action.

An individual can't be deleted if their profile is locked.

Solves #347 .


ui/src/components/IndividualsTable.stories.js
```diff
@@ -7,14 +7,22 @@ export default {
...
+`;
 
 export const Default = () => ({
   components: { IndividualsTable },
   template: IndividualsTableTemplate,
   methods: {
     queryIndividuals(page) {
       return this.query[page - 1];
+    },
+    deleteIndividual() {
```
sduenas, line 24: While testing this part, when I click to remove an entry, it doesn't remove anything from the table. Shouldn't it remove the entry from the table?

evamillan: I'm not sure if we should try to replicate everything that the backend does on Storybook, from what I've seen it's usually just used to show how the components look (although they are usually simpler components).

sduenas: My doubt is if to remove en entry is part of the logic of the component, that shouldn't be included on the component itself? Or do we expect to have different behaviours with the delete option depending on where the component is used?

evamillan: Do you mean that the delete function should be inside the component and not passed as a prop?

sduenas: I think that the functionality that removes the entry from the table, not from the server, should be part of it.  When I tested it in storybook I find odd that the entry is still on the table. I'm not sure, though. In any case, if you think it's better this way, we can leave it for now and think about it later.

evamillan: We could remove the entries from the table, but reloading the page after that to fill in the missing entries kind of "overwrites" it. I think if you find it OK and it works on the app, we can discuss it and improve the functionality on Storybook later.

sduenas: Ok! Make sense.

sduenas (approved):
The PR looks good to me. In any case, for next PRs I think it would be better if you can split the functionality in several commits for the same PR. For example, a commit with the apollo queries and their tests (delete queries), a next one with the implementation of the component (adding the delete functionality to the table), and a final one with the integration of that component into the App.

PR #353, opened 2020-10-16
[ui] Merge individuals on the workspace

evamillan:
Individuals in `WorkSpace` can be merged by clicking the merge icon on top of the component. When merged, the previously saved individuals get replaced by the merged one. Then the work space emits an event to indicate that other components may need to update their individuals.


sduenas (approved):
LGTM

PR #357, opened 2020-10-19
[ui] Merge individuals on the workspace using drag and drop 

evamillan:
Individuals can be merged by dragging them from the table and dropping them into an individual card in the workspace. The workspace is then updated with the new individual.

### DIFF
--- a/ui/src/apollo/mutations.js
+++ b/ui/src/apollo/mutations.js
@@ -41,6 +41,7 @@ const MERGE = gql`
         }
         profile {
           name
+          id
         }
         enrollments {
           organization {
--- a/ui/src/components/IndividualCard.vue
+++ b/ui/src/components/IndividualCard.vue
@@ -1,5 +1,14 @@
 <template>
-  <v-card class="mx-auto" raised v-on="$listeners">
+  <v-card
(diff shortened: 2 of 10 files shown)


sduenas (approved):
LGTM. I have a minor question, though.

Is it possible to have a different animation when dragging several individuals at once? It looks like I was only moving and merging one, not all the selected individuals.

evamillan:
You can only drag one individual at a time, just like in the table. Should you be able to move all the selected ones at once?

sduenas:
> You can only drag one individual at a time, just like in the table. Should you be able to move all the selected ones at once?

I think it should. Otherwise, users would feel odd that only the one dragged is merged when there are others selected for that operation.

evamillan:
That makes sense, but it would change the current logic of the drag and drop quite a bit. Should I change it here or make a different issue and PR for it?

sduenas:
> That makes sense, but it would change the current logic of the drag and drop quite a bit. Should I change it here or make a different issue and PR for it?

I think you can change it here. It's part of the requirements for this task.

evamillan:
I updated the PR.

PR #360, opened 2020-10-20
Show extended data on the IndividualCard

evamillan:
`IndividualCard` shows a compact version of `ExpandedIndividual` when the magnifier icon is clicked, which displays a list of the individual's identities and enrollments.


sduenas (approved):
LGTM

PR #363, opened 2020-10-21
Highlight Individuals that are on the workspace and on the table

evamillan:
Individuals on the workspace that are also visible on IndividualsTable display an icon and vice versa. When the user hovers over them, the corresponding individual is highlighted.


sduenas (approved):
LGTM

Issue #361, opened 2020-10-21
Restyle IndividualCard


sduenas:
We created a new design for the IndividualCard. The changes are not many, though. This is how the card should look like:

![imagen](https://user-images.githubusercontent.com/833352/96696739-66bd5c80-138b-11eb-9cc6-f5f5e58a4598.png)

sduenas:
Fixed by #364 .

PR #364, opened 2020-10-22
[ui] Restyle IndividualCard

evamillan:
Styles `IndividualCard` according to #361 , adding the individual's organization and replacing the lock button with one that removes the card from the workspace.


sduenas (approved):
LGTM

PR #371, opened 2020-10-26
[ui] Split identities on ExpandedIndividual

evamillan:
An individual's identity can be unmerged by clicking a button on the `ExpandedIndividual` identities list. The split individual is then placed on the workspace.
A new Apollo mutation is implemented to split the identities.


sduenas (approved):
LGTM

PR #374, opened 2020-10-27
Move Identity to another IndividualEntry or IndividualCard

evamillan:
An individual's identity on `ExpandedIndividual` can be dragged and dropped into an individual's card on the workspace or to another individual's entry on `IndividualsTable`.
A GraphQL mutation is added to allow an identity to be moved.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -51,6 +51,7 @@
             :delete-item="deleteItem"
             :merge-items="mergeItems"
             :unmerge-items="unmergeItems"
+            :move-item="moveItem"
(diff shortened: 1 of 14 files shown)


sduenas (changes requested):
The code is ok but I found a bug. When identities are moved, the info is not updated in the individual cards. Clicking on the + button shows the same info that was before the operation. The same happens when individuals are merged.

evamillan:
I updated the PR with the fix.

sduenas (approved):
LGTM

PR #375, opened 2020-10-28
[ui] Remove highlight when individual is not on the workspace

### LINKED ISSUE #373: Individual entries that are not in the workspace have the border highlighted
The border should only be highlighted for those Individuals that are in the workspace. To reproduce the error follow these steps:

1. Add an Individual from the table to the workspace
2. Remove the Individual from the workspace
3. Deselect the Individual from the table
4. The border will be highlighted  

selected

evamillan:
Stops highlighting an individual in the table when that individual has been removed from the workspace.
Fixes #373


sduenas (approved):
LGTM

PR #379, opened 2020-10-29
[ui] Show Organizations domains 

evamillan:
Adds the `ExpandedOrganization` component, which displays a list of an organization's domains. This component is shown when an entry on `OrganizationsTable` is expanded.


sduenas (approved):
LGTM

PR #382, opened 2020-10-30
[schema] Allow query individuals by any of their UUIDs

mafesan:
Before this change, the filter `uuid` from the Individuals query only worked when that `uuid` matched with the `mk` from the individual. Now, this `uuid` could be from any of its identities.

This PR is still WIP:
- [x] Update `Individuals` query filter.
- [x] Unit tests updated.


mafesan:
@sduenas I have rebased this branch with the latest changes from `muggle`, so this is now ready for review again. Thanks!

Issue #383, opened 2020-11-03
Add organizations from the UI


sduenas:
The user will be able to add a new organization by clicking `Add` button on the organizations table.
When the user clicks that button, a modal will show up. This modal contains a form where the user can introduce the name of the new organization and the different domains assigned to it.

Take the next mock-up as reference:

![imagen](https://user-images.githubusercontent.com/833352/98009400-c721b480-1df5-11eb-8004-2442dc522e9a.png)

sduenas:
Fixed by #386 .

PR #386, opened 2020-11-04
Add organizations from the UI 

evamillan:
Implements the `OrganizationModal` component, which displays a form that allows an organization and its domains to be created. The component is embedded inside `OrganizationsTable` and opens when the "add" button is clicked, as requested in  #383 . Two new GraphQL mutations -`addOrganization` and `addDomain`- are added to save the changes in the database.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -66,15 +66,11 @@
           />
         </v-col>
         <v-col class="organizations elevation-2">
-          <h4 class="title">
-            <v-icon color="primary" left>
-              mdi-sitemap
-            </v-icon>
-            Organizations
-          </h4>
           <organizations-table
             :fetch-page="getOrganizationsPage"
             :enroll="enroll"
+            :add-organization="addOrganization"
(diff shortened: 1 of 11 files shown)


sduenas (changes requested):
When a new organization is added with a domain that already exists, the app doesn't return any error. I think it should return an error.

evamillan:
I updated the PR. Now the errors are shown on an alert so that the user is aware of them and can fix them.

sduenas (changes requested):
When I add a domain that already exists I don't see any error on the app. Also, after adding it, when I click again to add a new organization, I can't write the name of the new org and also, the domain I tried to open before is on the form. You can see that behaviour in the next video:

error-new-org

evamillan:
I updated the PR. Since it was confusing, the "add" button is now a "+" and instead of adding the domain to a list, it adds a new input.

sduenas (approved):
LGTM

PR #387, opened 2020-11-10
Add Individuals from the UI

evamillan:
Adds the `ProfileModal` component, which displays a form that allows a new individual to be created. The component
uses three Apollo mutations, passed as props, to create the individual, update their profile information and enroll them in organizations. The modal is added inside `IndividualsTable` and can be opened clicking the "add" button.


sduenas (changes requested):
In general, the PR looks good to me but it needs some changes before we merged it:

- The only required argument in the form should be the `source` but the user should provide at least one of `name`, `email`, or `username` to create the new identity.
- The dates should be given using the format YYYY-MM-DD. This format doesn't create problems trying to understand where the month or the day goes. Also, I wonder if make sense to use a date picker for the ranges.
- For the country code, we can use a list of countries instead. Not sure if we can pre-loaded somehow but the ideal would be to load it from SortingHat.

evamillan:
I updated the PR with the changes you requested. Since the individual's name and email are not required, I had to change the cards and table entries a bit too so they weren't completely empty. 

imagen

sduenas:
> I updated the PR with the changes you requested. Since the individual's name and email are not required, I had to change the cards and table entries a bit too so they weren't completely empty.
> 
> imagen

Individuals `name` field will be always available because it's part of the profile. When a new individual is created, the server assigns a name based on the fields that were given. If no name was given, the value will be an email or a username but it will be stored under `name`.

Having said that, I don't think you need the part of the code where you treat that. You just need to take `name` form the profile and show it in the card.

ui/src/components/IndividualCard.vue
```diff
@@ -112,8 +117,12 @@ export default {
...
   computed: {
     getNameInitials: function() {
-      var names = this.name.split(" ");
-      var initials = names[0].substring(0, 1).toUpperCase();
+      const name =
+        this.name ||
+        this.identities[0].identities[0].username ||
+        this.identities[0].identities[0].email;
+      const names = name.split(" ");
+      let initials = names[0].substring(0, 1).toUpperCase();
```
sduenas, line 125: As I explained in my comment before, this isn't needed. The code should take the field `name` from the profile, not from the identities.

ui/src/components/IndividualEntry.vue
```diff
@@ -148,8 +152,9 @@ export default {
   }),
   computed: {
     getNameInitials: function() {
-      var names = this.name.split(" ");
-      var initials = names[0].substring(0, 1).toUpperCase();
+      const name = this.name || this.username || this.email;
+      const names = name.split(" ");
+      let initials = names[0].substring(0, 1).toUpperCase();
```
sduenas, line 157: Same comment as in the card component.

sduenas (changes requested):
The Individual can't be added if a country is not given. Also, check my other comments.

evamillan:
I fixed the error when no country is selected and changed the card and entry code according to your comments.

sduenas (approved):
LGTM

PR #388, opened 2020-11-11
Edit organizations from the UI

evamillan:
Adds an edit icon to `OrganizationEntry`, which opens the `OrganizationModal` component. `OrganizationTable` passes the organization information and a new `deleteDomain` Apollo mutation to the modal so that the domains can be edited.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -75,6 +75,7 @@
             :enroll="enroll"
             :add-organization="addOrganization"
             :add-domain="addDomain"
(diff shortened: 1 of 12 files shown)


sduenas (changes requested):
The PR looks good but I think we should allow to delete every domain not when there are only two or more.

evamillan:
I updated the PR.

sduenas (approved):
LGTM

PR #391, opened 2020-11-12
[ui] Allow all organizations in ProfileModal to be removed

evamillan:
All enrollments in `ProfileModal` can be deleted by clicking the trashcan button next to them, like the domains in `OrganizationModal`. The add organization button is placed below them.


sduenas (approved):
LGTM

PR #392, opened 2020-11-16
Add search function to the Individuals table

evamillan:
This PR adds an input to `IndividualsTable` that allows the user to search individuals by term, and an optional filters argument to the `getPaginatedIndividuals` Apollo query. The query is not performed if the search term is shorter than 3 characters. The input can be cleared to return the full list of individuals.


sduenas (approved):
LGTM.

Issue #390, opened 2020-11-12
Complete profile name with username value when name is not given


sduenas:
When a new Individual is created its profile is created at the same time. When the name value is not given, the `name` attribute on the profile will be empty. It would be better to fill it with the value of `username`, it this one is given.

mafesan:
@sduenas I have submitted the PR adding these changes, please have a look: https://github.com/chaoss/grimoirelab-sortinghat/pull/396. Thanks!

sduenas:
Fixed by #396.

PR #396, opened 2020-11-18
[api] Set username as profile name when no name is provided

mafesan:
As requested on issue #390, these changes allow to set the `username` as the profile name in case no `name` value is provided when a new Individual is created.


sduenas (approved):
LGTM

Issue #394, opened 2020-11-17
Search by last modified date


sduenas:
The goal of this task is to allow to search `Individuals` by the date when they were modified the last time. We can use something similar to what GitHub does with its [advanced search](https://docs.github.com/en/free-pro-team@latest/github/searching-for-information-on-github/understanding-the-search-syntax).

The filter should allow to search Individuals that were modified:

- [ ] on a given date
- [ ] before/on a date
- [ ] after/on a date
- [ ] between a range of dates

mafesan:
Hi @sduenas , I have submitted a draft PR with the main changes. I'm still working on the unit tests.

You will see the accepted date format is ISO 8601. I have included date + time, as other fields we have using dates need that format too. For instance, when adding an enrollment, the dates need to be following this format (e.g. `2008-01-01T00:00:00`).

mafesan:
PR https://github.com/chaoss/grimoirelab-sortinghat/pull/401 is now ready for review, thanks!

PR #401, opened 2020-11-23
[schema] Support filtering individuals by last updated date

mafesan:
Now, individuals can be queried filtering by their last updated date.

The accepted filter formats are controlled by regular expressions matching two patterns:
* A comparison operator (>, >=, <, <=) and a date (e.g. `>=YYYY-MM-DDTHH:MM:SSZ`).
* A range operator (..) between two dates (e.g. `YYYY-MM-DDTHH:MM:SSZ..YYYY-MM-DDTHH:MM:SSZ`)

The accepted date format is ISO 8601, `YYYY-MM-DDTHH:MM:SSZ`, also accepting microseconds and time zone offset (`YYYY-MM-DDTHH:MM:SS.ms+HH:HH`).

Note: This PR is linked to issue #394 .

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -21,6 +21,7 @@
 #
 
 import json
+import re
 
 import graphene
 import graphql_jwt
@@ -38,6 +39,9 @@ from graphene.types.generic import GenericScalar
 from graphene_django.converter import convert_django_field
 from graphene_django.types import DjangoObjectType
 
+from grimoirelab_toolkit.datetime import (str_to_datetime,
+                                          InvalidDateError)
+
 from .api import (add_identity,
                   delete_identity,
                   update_profile,
@@ -54,6 +58,7 @@ from .api import (add_identity,
                   withdraw)
 from .context import SortingHatContext
 from .decorators import check_auth
+from .errors import InvalidFilterError
 from .jobs import (affiliate,
                    unify,
                    find_job,
@@ -75,6 +80,56 @@ def convert_json_field_to_generic_scalar(field, registry=None):
     """Convert the content of a `JSONField` loading it as an object"""
 
     return OperationArgsType(description=field.help_text, required=not field.null)
+
+
+def parse_date_filter(filter_value):
+    """Extract the filter terms from a date filter
+
+    The accepted formats are controlled by regular expressions
+    matching two patterns: a comparison operator (>, >=, <, <=) and a date
+    OR a range operator (..) between two dates.
+
+    The accepted date format is ISO 8601, YYYY-MM-DDTHH:MM:SSZ, also
+    accepting microseconds and time zone offset (YYYY-MM-DDTHH:MM:SS.ms+HH:HH).
+
+    :param filter_value: String containing the filter value
+
+    :returns: A dictionary including an operator and the datetime values
+    """
+    # Accepted date format is ISO 8601, YYYY-MM-DDTHH:MM:SSZ (no `Z` is accepted too)
+    filter_data = {
+        "operator": None,
+        "date1": None,
+        "date2": None
+    }
+
+    iso_date_group = r"(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?(Z|\+\d{2}:\d{2})?)"
+    # Filter with a comparison operator (>, >=, <, <=) and a date (e.g. `>=YYYY-MM-DDTHH:MM:SS`)
+    oper_comparison = r"^(<=?|>=?)%s$" % iso_date_group
+    # Filter with a range operator (..) between two dates
+    # (e.g. YYYY-MM-DDTHH:MM:SSZ..YYYY-MM-DDTHH:MM:SSZ)
+    range_comparison = r"^%s\.{2}%s$" % (iso_date_group, iso_date_group)
+
+    oper_result = re.match(oper_comparison, filter_value)
+    range_result = re.match(range_comparison, filter_value)
+
+    if not oper_result and not range_result:
+        raise ValueError('Filter format is not valid')
+
+    if oper_result:
+        filter_data['operator'] = oper_result.group(1)
+        filter_data['date1'] = str_to_datetime(oper_result.group(2))
+
+    if range_result:
+        filter_data['operator'] = '..'
+        filter_data['date1'] = str_to_datetime(range_result.group(1))
+        filter_data['date2'] = str_to_datetime(range_result.group(4))
+
+        if filter_data['date1'] > filter_data['date2']:
+            range_msg = 'Date range is invalid. Upper bound must be greater than the lower bound'
+            raise ValueError(range_msg)
+
+    return filter_data
 
 
 class PaginationType(graphene.ObjectType):
@@ -198,6 +253,12 @@ class IdentityFilterType(graphene.InputObjectType):
     uuid = graphene.String(required=False)
     term = graphene.String(required=False)
     is_locked = graphene.Boolean(required=False)
+    last_updated = graphene.String(
+        required=False,
+        description='Filter with a comparison operator (>, >=, <, <=) and a date OR with a range operator (..) between\
+                     two dates, following ISO-8601 format. Examples:\n* `>=2020-10-12T09:35:06.13045+01:00` \
+                     \n * `2020-10-12T00:00:00..2020-11-22T00:00:00`.'
+    )
 
 
 class TransactionFilterType(graphene.InputObjectType):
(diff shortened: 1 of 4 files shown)


sortinghat/core/schema.py
```diff
@@ -77,6 +78,45 @@ def convert_json_field_to_generic_scalar(field, registry=None):
...
+    """
+    # Accepted date format is ISO 8601, YYYY-MM-DDTHH:MM:SS
+    filter_data = {
+        "operator": None,
+        "date1": None,
+        "date2": None
+    }
+
+    # Filter with a comparison operator (>, >=, <, <=) and a date (e.g. `>=YYYY-MM-DDTHH:MM:SS`)
+    oper_comparison = r"^(<=?|>=?)(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})$"
```
sduenas, line 102: This doesn't allow to look for an exact date. What would be the way to implement it?

sduenas (commented):
I have some comments we need to clarify before we continue.

sortinghat/core/schema.py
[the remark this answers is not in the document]
mafesan: You are right, but I chose to follow this format because it is the one that other date fields are using (like the enrollment dates in the `Enroll` mutation), as it is the one coming from the `DateTime` type from `graphene`.

mafesan: As the accepted format I chose included date+time, it didn't make sense to me adding the `==` operator. I think the way it is done now, it could be set as a range filter like `YYYY-MM-DDT00:00:00..YYYY-MM-DDT23:59:59`, WDYT?

tests/test_schema.py
```diff
@@ -1885,6 +1922,329 @@ def test_filter_non_exist_registry(self):
...
+        id1 = identities[0]
+        self.assertEqual(id1['email'], 'jsmith@example.com')
+
+        id2 = identities[1]
+        self.assertEqual(id2['email'], 'jsmith@bitergia.com')
+        self.assertEqual(id2['source'], 'scm')
+
+        id3 = identities[2]
+        self.assertEqual(id3['email'], 'jsmith@bitergia.com')
+        self.assertEqual(id3['source'], 'mls')
```
sduenas, line 2041: You can remove this. I don't think we need to check this. Checking the `mk` only should be enough.

sduenas (changes requested):
Please check my comments. The main problem is related to how to know when a filter is working and when it's not.

sortinghat/core/schema.py
```diff
@@ -77,6 +82,61 @@ def convert_json_field_to_generic_scalar(field, registry=None):
...
+    oper_comparison = r"^(<=?|>=?)%s$" % iso_date_group
+    # Filter with a range operator (..) between two dates
+    # (e.g. YYYY-MM-DDTHH:MM:SS..YYYY-MM-DDTHH:MM:SS)
+    range_comparison = r"^%s\.{2}%s$" % (iso_date_group, iso_date_group)
+
+    oper_result = re.match(oper_comparison, filter_value)
+    range_result = re.match(range_comparison, filter_value)
+
+    if not oper_result and not range_result:
+        raise InvalidFilterError(msg='Filter format is not valid')
```
sduenas, line 117: It would be better if you explain what the problem is and how someone can fix it.

sortinghat/core/schema.py
```diff
@@ -77,6 +82,61 @@ def convert_json_field_to_generic_scalar(field, registry=None):
...
+    if range_result:
+        filter_data['operator'] = '..'
+        try:
+            filter_data['date1'] = str_to_datetime(range_result.group(1))
+            filter_data['date2'] = str_to_datetime(range_result.group(4))
+        except InvalidDateError as e:
+            raise InvalidFilterError(msg=e)
+
+        if filter_data['date1'] > filter_data['date2']:
+            raise InvalidFilterError(msg='Date range is invalid, date 2 must be greater than date 1')
```
sduenas, line 135: Instead of saying `date 2` and `date 1`, why don't we say `Date range is invalid. Upper bound must be greater than the lower bound", or `Date range is invalid. First date must be greater than the second one". Let's try to find something clear.

sduenas (changes requested):
LGTM but some minor changes are needed.

sortinghat/core/schema.py
mafesan: In this case I think it is better to include some basic documentation within the field itself using this graphene structure when the filter is declared:
```
 graphene.String(
        required=False,
        description='Filter with a comparison operator (>, >=, <, <=) and a date OR with a range operator (..) between two dates. E.g. `>=YYYY-MM-DDTHH:MM:SS`.'
    )
```
About the information the error has to provide, the problem there is the filter did not match with any of the accepted regular expressions. Should this error print the not matching parts together with the expression that should match?

mafesan: This description would look like this in the API web documentation:

doc-last-updated-filter

@sduenas, WDYT ?

sduenas: I think it's good to have that in the documentation. However, one of the problems of the exception is that it doesn't says what filter is failing, so the user won't be able to know what to change.

mafesan: Oh, I see what you meant. The error was not telling `last_updated` is the filter field that is failing, I've just updated it catching the exception when the filter fails while the query is being resolved.

mafesan: BTW, I kept the piece of documentation I posted above with some real examples:
* `>=2020-10-12T09:35:06.13045+01:00` 
* `2020-10-12T00:00:00..2020-11-22T00:00:00`

mafesan:
Thank you so much for your comments @sduenas ! Now, ` parse_date_filter` method raises `ValueError` when the filter format is wrong or the date range is invalid; and `InvalidDateError` when the filter contains a wrong date. Both errors are captured when the query is being resolved, raising a `InvalidFilterError`, which gives information about the filter causing the problem and the cause of the error.

sduenas (changes requested):
The code looks ok. I've been testing it but it doesn't returns what it should, though.
For example, see the two screenshots. The first one shows the individuals that I have on the database. The second one shows a query that should return those individuals.

imagen

imagen

mafesan:
> The code looks ok. I've been testing it but it doesn't returns what it should, though.
> For example, see the two screenshots. The first one shows the individuals that I have on the database. The second one shows a query that should return those individuals.
> 
> imagen
> 
> imagen

You are right @sduenas, thank you for the heads-up. At least one of the individuals should be returned when filtering by that date... :thinking: 
I did test it and it worked, as I added the `__date__` section when filtering, as they explain in [Django documentation](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#date), but thinking about it, I added this part when I was taking the date as a string, and then Django casted that as a datetime object.

Now, the `str_to_datetime` method from GrimoireELK does this conversion already in the `parse_date_filter` method, so instead of calling the filter as `<filed_name>__date__<operator>` (`last_modified__date__gte`), it can be rewritten directly as `<field_name>__<operator>` (`last_modified__gte`), as the date is already a datetime object. After applying this change, it seems it works as expected (the unit tests worked in both cases, though).

Please, can you test again with the new changes? Thanks!

mafesan:
@sduenas I have modified the unit tests to cover the case you mentioned. 

The problem was related about how to build the Django query filter:  when building the query as `<filed_name>__date__<operator>`, the functionality was not working from the interface but the unit tests did not fail.

To fix this error, I modified how the Django query filter was built to `<field_name>__<operator>` but the tests kept working with no modification. Now I have added more cases to ensure this case is covered: the tests work with this updated filter and they fail if the filter is replaced by the former one.

evamillan:
@mafesan The JavaScript `toISOString` function returns a date with a `YYYY-MM-DDTHH:mm:ss.sssZ` format, but this kind of date raises a `Filter format is not valid` error when I test the PR. This is the format that we use for all the other dates, like the ones in the `enroll` mutation. Could you please change it?

mafesan:
> @mafesan The JavaScript `toISOString` function returns a date with a `YYYY-MM-DDTHH:mm:ss.sssZ` format, but this kind of date raises a `Filter format is not valid` error when I test the PR. This is the format that we use for all the other dates, like the ones in the `enroll` mutation. Could you please change it?

Thank you for the heads-up @evamillan! I have modified the accepted date formats to include the format you mentioned, I have tested it with the search box from the changes in #408 and it worked for me, can you please confirm it works for you too? Thanks!

evamillan:
I tested it with the last changes and it works. Thank you @mafesan !

sduenas (approved):
LGTM

PR #404, opened 2020-11-23
[ui] Remove individuals affiliations

evamillan:
This PR adds the `withdraw` Apollo mutation, which removes an individual's affiliation to an organization.
An individual's affiliation can be removed by clicking the delete button next to it on `ExpandedIndividual`.

### DIFF
--- a/ui/src/apollo/mutations.js
+++ b/ui/src/apollo/mutations.js
@@ -245,6 +245,51 @@ const DELETE_DOMAIN = gql`
   }
 `;
 
+const WITHDRAW = gql`
+  mutation withdraw(
+    $uuid: String!
+    $organization: String!
+    $fromDate: DateTime
+    $toDate: DateTime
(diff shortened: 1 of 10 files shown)


ui/src/apollo/mutations.js
```diff
@@ -384,6 +419,17 @@ const deleteDomain = (apollo, domain) => {
   return response;
 };
 
+const withdraw = (apollo, uuid, organization) => {
```
sduenas, line 422: This mutation must also get the period of time to remove. Remember that an Individual could be affiliated to the same organization during different periods of time.

sduenas (changes requested):
Please check my comment. The rest of the PR looks good to me.

evamillan:
Thank you, I updated the PR.

sduenas (approved):
LGTM

PR #405, opened 2020-11-23
Limit SQLAlchemy versions to <1.4

sduenas:
SQLAlchemy 1.4.x doesn't work with the current code in SortingHat. This is a hotfix to force pip to install a version lower than 1.4.

Signed-off-by: Santiago Dueñas <sduenas@bitergia.com>


zhquan (approved):
LGTM

PR #408, opened 2020-11-27
[ui] Individuals search box

evamillan:
This PR adds the `Search` component, which offers hints of the available filters and syntax to users as they type. So far, the available filters are `term` and `lastModified`. The search value is then parsed and emitted to the parent component `IndividualsTable` to perform the query.

Vuetify is upgraded to version ^2.3.19 to solve an issue with  the `v-combobox` component used on the search. Some components and CSS classes had been deprecated, so they have been updated.

Note: running a query with the lastModify filter will throw an error until the server part is ready.


evamillan:
I updated the PR. I added tests and fixed some issues on the Search component that I noticed while testing.

ui/tests/unit/queries.spec.js
```diff
@@ -335,60 +335,56 @@ describe("Search", () => {
...
-    ["lastUpdated", "range", "2000-2001"],
-    ["lastUpdated", "range", "2001-2000"]
-  ])("Given an invalid value renders an error", async (...args) => {
+    "lastUpdated:abc",
+    "lastUpdated:<abc",
+    "lastUpdated:<2000-23-01",
+    "lastUpdated:>=2000-01-49",
+    "lastUpdated:>@",
+    "lastUpdated:2000-2001",
+  ])("Given an invalid value renders an error", async (value) => {
```
sduenas, line 381: I think there must be a test case where `lastUpdated` has two operators. It should return an error too. For example `lastUpdated:>=2000-01-01<=2000-01-01` or `lastUpdated:===2000-01-01`.

ui/tests/unit/queries.spec.js
```diff
@@ -335,60 +335,56 @@ describe("Search", () => {
   };
 
   test.each([
```
sduenas, line 337: I think these tests should be in a different file. These are the tests for the queries.

sduenas (changes requested):
The PR looks good to me but there are some minor comments.

evamillan:
I added the test cases and moved them to a separate file.

sduenas (approved):
LGTM

PR #410, opened 2020-12-10
[ui] Remove organization

evamillan:
Adds a delete button to `OrganizationEntry` that removes it from the list using the `removeOrganization` Apollo GraphQL mutation. It displays a confirmation modal before the action is performed.

### DIFF
--- a/ui/src/apollo/mutations.js
+++ b/ui/src/apollo/mutations.js
@@ -290,6 +290,16 @@ const WITHDRAW = gql`
   }
 `;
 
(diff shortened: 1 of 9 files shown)


sduenas (changes requested):
I've been testing the PR and I found a little problem. When an organization is removed and there's an individual listed with that organization, the enrollment doesn't disappear unless I refresh the page. Not sure if we can do something to fix this problem.

evamillan:
I updated the PR.

sduenas (approved):
LGTM

PR #412, opened 2020-12-15
[ui] Drag and drop organizations to enroll individuals

evamillan:
An organization can be dragged and dropped into an `IndividualCard` or an `IndividualEntry`. They recover the organization's name and pass it to their parent component, `IndividualsTable` or `Workspace`. The parent then uses the functions in `enrollMixin` to enroll the individual.


sduenas (approved):
LGTM

PR #415, opened 2020-12-16
[ui] Fix removed cards reappearing in workspace

### LINKED ISSUE #414: [ui] Deleted cards in workspace reappear when an identity is split
When an identity is split from an individual, it is automatically moved to the workspace. When this happens, any card that had previously been removed from the workspace reappears.

evamillan:
Fixes #414. When individual cards were removed from the workspace, the list of individuals in the workspace that was saved in the dashboard was not updated. This made them reappear when an identity was split from an individual and its card was moved to the workspace.

### DIFF
--- a/ui/src/apollo/mutations.js
+++ b/ui/src/apollo/mutations.js
@@ -59,6 +59,8 @@ const MERGE = gql`
           organization {
             name
           }
+          start
+          end
         }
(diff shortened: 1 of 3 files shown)


sduenas (approved):
LGTM

PR #416, opened 2020-12-16
[ui] Show dasource name on datasource icons

evamillan:
Adds a tooltip that shows the datasource name when the cursor is over an icon.


sduenas (approved):
LGTM

PR #420, opened 2020-12-18
[ui] Show icons for known datasources

evamillan:
Adds icons for the following datasources:

- dockerhub
- jira
- rss
- slack
- stackexchange
- telegram
- twitter

Some of the icon names didn't match their datasource name and the order of the icons changed, so it updates the  `groupIdentities` function to add both the datasource and the icon to the list, and to sort them in alphabetical order with the "other sources" always last.


ui/src/utils/actions.js
```diff
@@ -36,24 +36,51 @@ const moveIdentity = (fromUuid, toUuid, action, dialog) => {
...
-  return Object.values(groupedIdentities);
+  // Sort identities by alphabetical order
+  let sortedIdentities = Object.values(groupedIdentities).sort((a, b) =>
+    a.name.localeCompare(b.name)
+  );
+
+  // Move "other" identities to end of list
+  sortedIdentities.push(
+    ...sortedIdentities.splice(
+      sortedIdentities.findIndex(identity => identity.name == "Other sources"),
```
sduenas, line 78: "Other sources" is used at least twice in the code. I think it would be better to assign it to a constant and reference the constant instead of the string.

sduenas (changes requested):
LGTM but with a minor comment to fix.

evamillan:
I changed it, thanks.

sduenas (approved):
LGTM

PR #424, opened 2020-12-28
[ui] Use GrimoireLab color scheme

evamillan:
Sets the blue and yellow GrimoireLab colors as primary and secondary colors in the Vuetify theme and changes some styles.

Note: the JWT authentication plugin is having issues at the moment: https://github.com/flavors/django-graphql-jwt/issues/241
If you can't log in, you can remove `meta: { requiresAuth: true }` from `/ui/src/router/index.js` to access the dashboard.

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -43,7 +43,8 @@ export default {
   }
 };
 </script>
-<style scoped>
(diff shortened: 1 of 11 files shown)


sduenas (approved):
LGTM

Issue #409, opened 2020-12-04
Mutation for updating enrollments


mafesan:
The goal of this task is to provide a mutation which allow users to update enrollment dates ensuring this operation is atomic. This mutation is related to issue #403 .

sduenas:
Fixed by #427.

PR #427, opened 2021-01-12
Add mutation for updating enrollment dates

mafesan:
This mutation updates atomically an individual's enrollment or set of enrollments defined by the initial date range to a new date range.

As in the enroll method wider date ranges get merged into the wider one, this mutation and its corresponding API method uses the `withdraw` and the `enroll` API methods to ensure the dates are updated even in this case following these steps:
 * Withdraw the enrollments from the provided organization and an existing date range.
 * Set a new enrollment to that same organization with the new date range.
 
By default, the `force` parameter from the `enroll` method is also included and set to `True` by default, as it overwrites default dates in case a more specific date is provided.

In case any of the new dates are missing, the former value for that date will be preserved. If both new dates are missing, the method will raise an `InvalidValueError`.

Some other minor changes are included in this PR: 
There were some log error messages which were not aligned with the format that other methods are using. The format consists on  surrounding with simple quotes the name of the field which the error is referred to.
For instance, the field `from_date` is referred as `'from_date'`.

Note: this PR is related with issue #409 .

Still WIP:
A final review from my side is still needed before marking this PR as ready for review.

### DIFF
--- a/sortinghat/core/api.py
+++ b/sortinghat/core/api.py
@@ -656,13 +656,13 @@ def enroll(ctx, uuid, organization, from_date=None, to_date=None, force=False):
         already exists in the registry.
     """
     if uuid is None:
-        raise InvalidValueError(msg="uuid cannot be None")
+        raise InvalidValueError(msg="'uuid' cannot be None")
     if uuid == '':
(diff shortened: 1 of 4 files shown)


sduenas (approved):
LGTM

PR #428, opened 2021-01-14
[CI] Migrate from Travis CI to GitHub Actions

vchrombie:
Travis CI has brought a few changes in its plans which is causing friction to the current workflow.

This commit removes the support of Travis CI and adds the support of GitHub Actions as the new CI service.


sduenas (approved):
LGTM

sduenas (commented):
Actions should only be run when there are commits or PR to `master` branch. `muggle` branch is unstable and this won't work if there are commits merged on it.

I'm not sure if this will run the actions on those cases but if it works like Travis, I'm afraid it will cause problems. Would you check it, please?

vchrombie:
Hi @sduenas 

> Actions should only be run when there are commits or PR to `master` branch. `muggle` branch is unstable and this won't work if there are commits merged on it.

Okay, it makes sense to me.

> I'm not sure if this will run the actions on those cases but if it works like Travis, I'm afraid it will cause problems. Would you check it, please?

The current actions workflow will be triggered for every PR or push to any branch. I understand that I need to change it. I will update it right now.

sduenas (approved):
LGTM

PR #429, opened 2021-01-18
[ui] Edit enrollment dates

evamillan:
This PR allows users to modify the dates of an individual's enrollment from the UI, using the `updateEnrollment` Apollo mutation.


sduenas (approved):
LGTM

PR #433, opened 2021-01-28
[ui] Reduce the space in between Individuals and Orgs tables

evamillan:
This PR makes the Individuals and Organizations containers fluid so the space between them is reduced and consistent in different screen sizes.


sduenas (approved):
LGTM

PR #434, opened 2021-01-28
[ui] Expand individual entry on double click

### LINKED ISSUE #431: Individual entry does not expand on double click
When double-clicking on an individual entry on the Individuals table, this entry does not expand to show the information. Organization entry expands when double-clicking.

evamillan:
Expands an individual's information when double clicking an entry. The select event performed on click is delayed for some milliseconds to prevent it from getting triggered on double click.

Fixes #431.

### DIFF
--- a/ui/src/components/IndividualEntry.vue
+++ b/ui/src/components/IndividualEntry.vue
@@ -7,7 +7,8 @@
       dropzone: isDropZone,
       highlighted: isHighlighted
     }"
-    @click="selectEntry"
+    @click.prevent="selectEntry"
+    @dblclick="onDoubleClick"
     @drop.stop="onDrop($event)"
     @dragover.prevent="isDragging = true"
     @dragenter.prevent="isDragging = true"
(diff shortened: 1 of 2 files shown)


ui/src/components/IndividualEntry.vue
```diff
@@ -223,10 +225,16 @@ export default {
   },
   methods: {
     selectEntry() {
-      if (this.isLocked) {
-        return;
+      if (!this.isLocked && !this.timeout) {
+        this.timeout = window.setTimeout(() => {
+          this.timeout = null;
+          this.$emit("select");
+        }, 350);
```
sduenas, line 232: I guess this is the timeone, isn't it? If that's the case, it would be better to have it under a constant and not like a magic number: https://en.wikipedia.org/wiki/Magic_number_(programming).

sduenas (changes requested):
The PR looks good. Just check a minor comment.

ui/src/components/IndividualEntry.vue
evamillan: You're right. I fixed it.

sduenas (approved):
LGTM

PR #437, opened 2021-02-01
[ui] Support Gravatar 

evamillan:
This PR adds support for Gravatar, which loads an individual's avatar from a URL created from their email. If the avatar does not exist, it returns a transparent image, which allows the default avatar with the individual's initials to be seen.
An md5 library is installed in order to create the hash for the Gravatar URL.


sduenas (changes requested):
The PR looks great but there are some minor issues:

1. I see a red border when I use it with my avatar. See the image below. Apparently is the default image with my initials that is under the avatar what is causing it.
imagen

2. I think we need two stories that show a real avatar. One for `IndividualEntry` and another for `IndividualCard`. You can use my avatar if you need it.

evamillan:
I think I've fixed the border around the avatar by making the image slightly bigger, but let me know if you can still see it. I also used you email for the Gravatar examples on Storybook, if that's OK.

sduenas (approved):
LGTM

PR #440, opened 2021-02-03
[ui] Make Workspace height fixed

### LINKED ISSUE #438: Empty workspace size changes when a new card is added
The size of the workspace when is empty changes when a new card is added to it. This creates a weird effect (see the capture below). I suggest to have a fixed size for the workspace.

workspace

evamillan:
Makes the empty Workspace have the same height as if it had a row of cards by default. Fixes #438.


sduenas:
LGTM

PR #441, opened 2021-02-04
Search organizations by term

evamillan:
The schema and UI are updated to allow filtering organizations by term:
- A `term` filter is added to the `organizations` query in `schema.py`, which returns a list of organizations whose name or domains contain the given term.
- Search filters are added to the `getPaginatedOrganizations` Apollo GraphQL query. The functions that call that query are updated accordingly.
- The `Search` component is imported in `OrganizationsTable`.

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -248,6 +248,7 @@ class CountryFilterType(graphene.InputObjectType):
 
 class OrganizationFilterType(graphene.InputObjectType):
     name = graphene.String(required=False)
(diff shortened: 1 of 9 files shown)


sduenas (changes requested):
In overall, the patch looks good but I have some comments:

- Add a test for a term that doesn't exist. The test should check that the returned organizations are zero.
- As in the individuals search box, the search will be requested when the number of characters is 3 or more.

sduenas (approved):
LGTM

PR #445, opened 2021-02-16
[schema] Add query to get the list of jobs

evamillan:
This PR adds the `jobs` query, which returns a paginated list of all the jobs in the main queue.

The query has the following format:

```
query {
  jobs(pageSize: Int, page: Int) {
    entities {
      jobId
      status
      jobType
      errors
      result
    }
    pageInfo {
      page
      pagesize
      numPages
      hasNext
      hasPrev
      totalResults
    }
  }
}
```

### DIFF
--- a/sortinghat/core/jobs.py
+++ b/sortinghat/core/jobs.py
@@ -59,6 +59,41 @@ def find_job(job_id):
 
     return jobs[0]
 
+def get_jobs():
+    """Get a list of all jobs
+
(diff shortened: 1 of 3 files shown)


sduenas (changes requested):
The patch looks ok. I've just only miss a test to check the pagination. Look for `test_pagination` within the file to get some examples.

evamillan:
@sduenas I've added a test to check the pagination.

PR #449, opened 2021-02-22
[ui] Save Workspace

evamillan:
The workspace data is saved on the Vuex store and local storage to prevent it from getting lost when the browser is closed or reloaded.


sduenas (approved):
LGTM

PR #461, opened 2021-02-24
Search bot individuals

evamillan:
Adds the `isBot` filter to the individuals query, which accepts a boolean value and returns a list of individuals filtered by their bot status.

### DIFF
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -352,6 +352,16 @@ SH_INDIVIDUALS_BOT_FILTER = """{
     }
   }
 }"""
+SH_INDIVIDUALS_BOT_FILTER_FALSE = """{
+  individuals(filters: {isBot: false}) {
+    entities {
+      mk
+      profile {
+        isBot
+      }
+    }
+  }
+}"""
 SH_INDIVIDUALS_LAST_UPDATED_FILTER = """{
   individuals(filters: {lastUpdated: "%s"}) {
     entities {
@@ -2037,6 +2047,18 @@ class TestQueryIndividuals(django.test.TestCase):
         indv = individuals[0]
         self.assertEqual(indv['mk'], 'a9b403e150dd4af8953a52a4bb841051e4b705d9')
         self.assertEqual(indv['profile']['isBot'], True)
+
+        # Test isBot: false
+        client = graphene.test.Client(schema)
+        executed = client.execute(SH_INDIVIDUALS_BOT_FILTER_FALSE,
+                                  context_value=self.context_value)
+
+        individuals = executed['data']['individuals']['entities']
+        self.assertEqual(len(individuals), 1)
+
+        indv = individuals[0]
+        self.assertEqual(indv['mk'], 'c6d2504fde0e34b78a185c4b709e5442d045451c')
+        self.assertEqual(indv['profile']['isBot'], False)
 
 
     def test_filter_non_exist_registry(self):
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -342,6 +342,16 @@ SH_INDIVIDUALS_LOCKED_FILTER = """{
     }
   }
 }"""
+SH_INDIVIDUALS_BOT_FILTER = """{
+  individuals(filters: {isBot: true}) {
+    entities {
+      mk
+      profile {
+        isBot
+      }
+    }
+  }
+}"""
 SH_INDIVIDUALS_LAST_UPDATED_FILTER = """{
   individuals(filters: {lastUpdated: "%s"}) {
     entities {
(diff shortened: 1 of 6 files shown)


tests/test_schema.py
```diff
@@ -1991,6 +2023,116 @@ def test_filter_registry_is_locked(self):
...
+                                email='jsmith@bitergia.com',
+                                username=None,
+                                source='mls',
+                                individual=indv)
+        Enrollment.objects.create(individual=indv, organization=org_ex)
+        Enrollment.objects.create(individual=indv, organization=org_bit,
+                                  start=datetime.datetime(1999, 1, 1, 0, 0, 0,
+                                                          tzinfo=dateutil.tz.tzutc()),
+                                  end=datetime.datetime(2000, 1, 1, 0, 0, 0,
+                                                        tzinfo=dateutil.tz.tzutc()))
```
sduenas, line 2067: We can remove some data that is not really needed on the test to make it easier to read.

tests/test_schema.py
```diff
@@ -1991,6 +2023,116 @@ def test_filter_registry_is_locked(self):
...
+                                email='jdoe@example.com',
+                                username='jdoe',
+                                source='scm',
+                                individual=indv)
+        Identity.objects.create(uuid='B002',
+                                name=None,
+                                email='jdoe@libresoft.es',
+                                username=None,
+                                source='scm',
+                                individual=indv)
```
sduenas, line 2086: This can be removed too.

tests/test_schema.py
```diff
@@ -1991,6 +2023,116 @@ def test_filter_registry_is_locked(self):
...
+
+        rol1 = enrollments[0]
+        self.assertEqual(rol1['organization']['name'], 'Bitergia')
+        self.assertEqual(rol1['start'], '1999-01-01T00:00:00+00:00')
+        self.assertEqual(rol1['end'], '2000-01-01T00:00:00+00:00')
+
+        rol2 = enrollments[1]
+        self.assertEqual(rol2['organization']['name'], 'Example')
+        self.assertEqual(rol2['start'], '1900-01-01T00:00:00+00:00')
+        self.assertEqual(rol2['end'], '2100-01-01T00:00:00+00:00')
```
sduenas, line 2134: This is not needed for the test.

tests/test_schema.py
```diff
@@ -342,6 +342,38 @@
     }
   }
 }"""
+SH_INDIVIDUALS_BOT_FILTER = """{
```
sduenas, line 345: We should reduce the data that is requested in this query to make the test easier to read and understand it. We only need `mk` and `isBot` fields.

ui/src/components/Search.vue
```diff
@@ -62,6 +62,8 @@ export default {
           const [filter, text] = value.split(":");
           if (filter === "lastUpdated") {
             this.parseLastUpdated(text);
+          } else if (text === "false") {
```
sduenas, line 65: What happens if the `text` is `"true"`?

sduenas (changes requested):
Please check my review.

evamillan:
@sduenas Thanks for the feedback. I removed the unnecessary data and added a case for when the search text is "true".

tests/test_schema.py
```diff
@@ -342,6 +342,16 @@
...
+  individuals(filters: {isBot: true}) {
```
sduenas, line 346: We should also test this for a `false` case.

sduenas (changes requested):
I've been testing this patch using the UI and I don't get any error when I set invalid values for `isBot`. It always returns identities that are not bots. On the other hand, if I use the graphql API I get errors.

```
query individuals {
  individuals(filters: { lastUpdated:">=2020-11-25T18:04:38.000Z", isBot:"falsee" }) {
    entities {
      mk
      lastModified
    }
  }
}

{
  "errors": [
    {
      "message": "Argument \"filters\" has invalid value {lastUpdated: \">=2020-11-25T18:04:38.000Z\", isBot: \"falsee\"}.\nIn field \"isBot\": Expected type \"Boolean\", found \"falsee\".",
      "locations": [
        {
          "line": 2,
          "column": 24
        }
      ],
      "extensions": {
        "code": 128
      }
    }
  ]
}
```

Can you check it, please?

evamillan:
> I've been testing this patch using the UI and I don't get any error when I set invalid values for `isBot`. It always returns identities that are not bots. On the other hand, if I use the graphql API I get errors.

@sduenas It looks like it converts the string to a boolean automatically when you query from the UI, but I can't find why it does that or how to prevent it. I've changed it so that it shows an error on the UI if the value is invalid and stops it from searching. Does this solution work for you?

sduenas:
> > I've been testing this patch using the UI and I don't get any error when I set invalid values for `isBot`. It always returns identities that are not bots. On the other hand, if I use the graphql API I get errors.
> 
> @sduenas It looks like it converts the string to a boolean automatically when you query from the UI, but I can't find why it does that or how to prevent it. I've changed it so that it shows an error on the UI if the value is invalid and stops it from searching. Does this solution work for you?

I've been tracing the source of the problem. Python `graphene` doesn't parse correctly `string` values for `Boolean` fields. It uses `bool(value)` to cast to a boolean value but this is wrong because it always returns `True` for string values.
Your proposed solution is valid for us.

sduenas (approved):

PR #463, opened 2021-02-24
Search individuals by country

evamillan:
Adds the `country` filter to the `individuals` query, which returns individuals with a given country name or country ISO codes.
In the UI `Search` component, the syntax is `country:"country name"`.

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -259,6 +259,7 @@ class IdentityFilterType(graphene.InputObjectType):
     is_locked = graphene.Boolean(required=False)
     is_bot = graphene.Boolean(required=False)
     gender = graphene.String(required=False)
+    country = graphene.String(required=False)
     last_updated = graphene.String(
         required=False,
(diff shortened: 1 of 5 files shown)


sduenas (changes requested):
LGTM but, would you rebase it so you don't need to add the same code that was added in the gender PR?

evamillan:
Done!

ui/src/components/Search.vue
```diff
@@ -68,7 +68,7 @@ export default {
           } else if (booleanFilters.find(bfilter => bfilter === filter)) {
             this.parseBooleanFilter(filter, text);
           } else {
-            this.filters[filter] = text;
+            this.filters[filter] = text.split("_").join(" ");
```
sduenas, line 71: Do we still need this?

evamillan: We don't. I've removed it.

sduenas (approved):
LGTM

PR #464, opened 2021-02-25
Search individuals by gender

evamillan:
Adds the `gender` filter to the` individuals` query, which returns individuals with a given gender.
In the UI Search component, the syntax is `gender:string`.


sduenas (commented):
Instead of using underscore characters to set spaces, wouldn't be better to use double quotes? It's a syntax widely used. For example, `gender:"non binary"`.

evamillan:
@sduenas  I changed it so that it uses double quotes instead of underscores.

sduenas (approved):
LGTM

PR #467, opened 2021-03-02
[ui] Select all individuals for batch operations

evamillan:
Adds a checkbox that selects or deselects all individuals on `IndividualsTable` to perform batch operations. 

- Locked individuals can be selected, but they still can't be merged or deleted.
- The action buttons and search boxes on both individuals and organizations tables are moved to a new row to make space for it, so the search box doesn't need to be closed by default.
-  New buttons to split all identities and to remove all affiliations are also added to `ExpandedIndividuals`. They are disabled for locked individuals.

This is a preview of the changes:
imagen

### DIFF
--- a/ui/src/components/ExpandedIndividual.vue
+++ b/ui/src/components/ExpandedIndividual.vue
@@ -47,7 +47,20 @@
         </v-edit-dialog>
       </div>
     </v-row>
-    <v-subheader>Identities ({{ identitiesCount }})</v-subheader>
+    <v-subheader class="d-flex justify-space-between">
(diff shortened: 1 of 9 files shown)


sduenas (changes requested):
The PR is ok but theres a minor problem. When the select box is checked and after performing an operation, the select box remains checked. I think it should be.

evamillan:
I've changed it so that if an action ends up deselecting individuals, like merging them or moving them to the workspace, the box stops being checked. I also rebased the branch.

sduenas (approved):
LGTM

PR #469, opened 2021-03-04
[ui] Show errors for invalid search filters

evamillan:
Adds the `validFilters` and `showFilters` props to the `Search` component.
- If the user input includes a filter that is not found on the valid filters list, an error is shown.
- With the `showFilters` prop, a dropdown with all valid filters is displayed next to the search box. When a filter is clicked, it is added to the search input.

### DIFF
--- a/ui/src/components/OrganizationsTable.vue
+++ b/ui/src/components/OrganizationsTable.vue
@@ -16,9 +16,15 @@
         Add
       </v-btn>
     </v-row>
+
     <v-row class="actions">
-      <search class="ma-0 mr-3 pa-0 flex-grow-0" @search="filterSearch" />
+      <search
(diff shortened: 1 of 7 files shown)


sduenas (changes requested):
These are my comments about the PR:

- I'd rather have the filters selector in a different PR. Different functionality, different PRs.
- When a filter that doesn't exist has a value in between double quotes, the error is not shown. For example `filter:"value"` will show nothing. Tests should be updated too.
- Boolean filters, such as `isBot` should show boolean values instead of `search value` when they are selected from the filters dropdown.
- I wonder if it's possible to have a better integration of the search box with the filters selector. Like in GitHub. Now, it looks it doesn't belong to the search box, the font is also bigger and different.

sduenas (approved):
LGTM

PR #470, opened 2021-03-09
[ui] Make table items per page editable

evamillan:
Adds an input to the `IndividualsTable` and `OrganizationsTable` footer where users can select how many items per page are shown.


evamillan:
I've rebased it.

PR #476, opened 2021-03-11
[ui] Add filter selector to the search box

evamillan:
Adds an optional selector that displays the list of valid filters next to the search input. When a filter is selected from the list, it is added to the input value.

### DIFF
--- a/ui/src/components/Search.vue
+++ b/ui/src/components/Search.vue
@@ -2,27 +2,57 @@
   <v-text-field
     v-model.trim="inputValue"
     append-outer-icon="mdi-magnify"
-    class="search ma-0 pa-0 flex-grow-0"
+    class="search pa-0 flex-grow-0"
     :error-messages="errorMessage"
     clearable
     label="Search"
     type="text"
+    height="30"
     hint=" "
     persistent-hint
+    single-line
+    outlined
+    dense
     @click:append-outer="search"
     @click:clear="clear"
     @keyup.enter="search"
     @focus="isFocused = true"
     @blur="isFocused = false"
   >
-    <template v-slot:message="{ key, message }">
-      <span>{{ message }}</span>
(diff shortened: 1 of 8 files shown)


ui/src/components/Search.vue
```diff
@@ -175,23 +209,47 @@ export default {
         vfilter => vfilter.filter === filter
       );
       return validFilter.type === "boolean";
+    },
+    setFilter(item) {
```
sduenas, line 213: We need some unit tests to check that when any the possible filters are clicked, the input is correctly set.

ui/src/components/Search.vue
```diff
@@ -9,20 +9,49 @@
...
   >
-    <template v-slot:message="{ key, message }">
-      <span>{{ message }}</span>
-      <v-btn href="/search-help" target="_blank" color="primary" text x-small>
-        View search syntax
-        <v-icon x-small right>
-          mdi-help-circle-outline
-        </v-icon>
-      </v-btn>
+    <template v-slot:prepend v-if="filterSelector">
```
sduenas, line 21: Can we improve the style of the search box?. In my opinion, it's very big if you compare it with other components of the page. In fact its size is the same or similar of the `Indiviuals` and `Organizations` table. Maybe reducing the size and the font might work.

sduenas (changes requested):
In overall works great but I think we need to improve some parts. What do you think?

ui/src/components/Search.vue
evamillan: I added the unit tests on `search.spec.js`.

ui/src/components/Search.vue
evamillan: I reduced the font and the overall size of the box a bit, let me know if I can do anything else to improve it further.

sduenas (approved):
LGTM

PR #479, opened 2021-03-15
[flake8] Fix flake8 errors

vchrombie:
This PR fixes the flake8 errors present in the muggle branch. This is necessary as a flake8 check must be added in the github action ci workflow.

It adds a configuration to ignore some errors of a file as it is hard to understand the logic if we format the file. It also adds a configuration to excludes a file.

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -335,6 +335,7 @@ class OperationPaginatedType(AbstractPaginatedType):
     entities = graphene.List(OperationType)
     page_info = graphene.Field(PaginationType)
 
+
 class JobPaginatedType(AbstractPaginatedType):
     entities = graphene.List(JobType)
     page_info = graphene.Field(PaginationType)
--- a/config/settings/devel.py
+++ b/config/settings/devel.py
@@ -114,7 +114,7 @@ DATABASES = {
     }
 }
 
-#FIXTURE_DIRS = "sortinghat/core/fixtures/"
+# FIXTURE_DIRS = "sortinghat/core/fixtures/"
 
 USE_TZ = True
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,6 @@
 description-file = README.md
 
 [flake8]
-exclude = .git, .eggs, __pycache__, build, dist, docs, docker, migrations
-ignore = E129, E402, F841
+exclude = .git, .eggs, __pycache__, build, dist, docs, docker, migrations, sortinghat/cli/client/schema.py
+ignore = E129, E402, F841, W504 E128 sortinghat/core/schema.py
 max-line-length = 130
--- a/setup.py
+++ b/setup.py
@@ -23,17 +23,13 @@
 
 import codecs
 import os.path
-import re
-import sys
-import unittest
 
 # Always prefer setuptools over distutils
 from setuptools import setup
-from setuptools.command.test import test as TestClass
 
 here = os.path.abspath(os.path.dirname(__file__))
 readme_md = os.path.join(here, 'README.md')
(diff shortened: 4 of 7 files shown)


vchrombie:
@sduenas, please review when you are free.
Thanks.

sortinghat/cli/client/schema.py
```diff
@@ -183,7 +183,8 @@ class CountryType(sgqlc.types.Type):
     code = sgqlc.types.Field(sgqlc.types.non_null(String), graphql_name='code')
```
sduenas, line 183: We should exclude this file from the checks. It's going to keep failing because it is generated automatically.

sortinghat/core/schema.py
```diff
@@ -826,10 +827,8 @@ def resolve_organizations(self, info, filters=None,
             query = query.filter(name=filters['name'])
         if filters and 'term' in filters:
             search_term = filters['term']
-            query = query.filter(Q(name__icontains=search_term) |
-                                 Q(name__in=Subquery(Domain.objects
-                                                    .filter(domain__icontains=search_term)
-                                                    .values_list('organization__name'))))
+            query = query.filter(Q(name__icontains=search_term) | Q(name__in=Subquery(
```
sduenas, line 830: I rather keep this as in the original file. The query is easier to ready. Is there anyway to avoid this problem?

sortinghat/core/schema.py
```diff
@@ -845,20 +844,15 @@ def resolve_individuals(self, info, filters=None,
...
             search_term = filters['term']
             # Filter matching individuals by their mk and their identities
-            query = query.filter(mk__in=Subquery(Identity.objects
-                                                 .filter(Q(name__icontains=search_term) |
-                                                         Q(email__icontains=search_term) |
-                                                         Q(username__icontains=search_term) |
-                                                         Q(individual__profile__name__icontains=search_term) |
-                                                         Q(individual__profile__email__icontains=search_term))
-                                                 .values_list('individual__mk')))
+            query = query.filter(mk__in=Subquery(Identity.objects.filter(
```
sduenas, line 852: Same as before.

sduenas (changes requested):
Please check my comments.

sortinghat/cli/client/schema.py
vchrombie: Okay. :+1:
The flake8 configuration file is missing anyway. I think we can add this file to the `exclude` list.

sortinghat/core/schema.py
vchrombie: Yes, we can do it. The flake8 documentation has a mention about it.
https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

vchrombie:
Hi @sduenas, I updated the PR accordingly.
- added the file `sortinghat/cli/client/schema.py` file to exclude list
- added `# noqa: code` to the specific lines mentioned

sduenas:
> Hi @sduenas, I updated the PR accordingly.
> 
>     * added the file `sortinghat/cli/client/schema.py` file to exclude list
> 
>     * added `# noqa: code` to the specific lines mentioned

I really hate to have those comments around to avoid these problems. Maybe we should exclude this file too, to be honest.

vchrombie:
> I really hate to have those comments around to avoid these problems. Maybe we should exclude this file too, to be honest.

I had the same opinion but did it because the errors might pop in some other place. Well, I think we need not be that strict. I will add the file to the ignore list.

vchrombie:
I rebased the PR to include the latest changes. The build works too! :tada:

sduenas (approved):
LGTM

PR #482, opened 2021-03-16
[ui] Add logging messages

evamillan:
Logs a message to the console when the following operations are performed, including the values that were changed:
- Add identity
- Move identity
- Update profile
- Enroll in organization
- Update enrollment
- Remove enrollment
- Merge identities
- Split identities
- Delete identity
- Add organization
- Add domain
- Remove domain
- Log in/out

The logs are not shown in production mode (when running `yarn serve --mode production`, for example).

### DIFF
--- a/ui/src/App.vue
+++ b/ui/src/App.vue
@@ -51,6 +51,7 @@ export default {
       Cookies.remove("sh_authtoken");
       Cookies.remove("sh_user");
       this.$router.push("/login");
+      this.$logger.info(`Log out user ${this.user}`);
     }
   }
 };
--- a/ui/src/components/IndividualsTable.vue
+++ b/ui/src/components/IndividualsTable.vue
@@ -364,6 +364,10 @@ export default {
         individuals.map(individual => this.deleteItem(individual.uuid))
       );
       if (response) {
+        this.$logger.debug(
+          "Deleted individuals",
+          individuals.map(individual => individual.uuid)
+        );
         this.queryIndividuals(this.page);
         this.dialog.open = false;
       }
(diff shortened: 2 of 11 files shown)


ui/src/App.vue
```diff
@@ -51,6 +51,7 @@ export default {
       Cookies.remove("sh_authtoken");
       Cookies.remove("sh_user");
       this.$router.push("/login");
+      this.$logger.log(`Log out user ${this.user}`);
```
sduenas, line 54: Set is as `info`.

ui/src/components/IndividualsTable.vue
```diff
@@ -359,6 +359,10 @@ export default {
         individuals.map(individual => this.deleteItem(individual.uuid))
       );
       if (response) {
+        this.$logger.log(
```
sduenas, line 362: Set it as `debug` message.

ui/src/plugins/logger.js
```diff
@@ -0,0 +1,18 @@
+export default {
+  install(Vue) {
+    function log(type, message, ...args) {
+      if (process.env.NODE_ENV === "production") {
+        return;
```
sduenas, line 5: I think for production we might want all messages but `debug` ones. If the app is reporting an error, users can give us useful information.

ui/src/plugins/logger.js
```diff
@@ -0,0 +1,18 @@
...
+      }
+      console[type](message, ...args);
+    }
+
+    Vue.prototype.$logger = {
```
sduenas, line 10: We need some unit tests for this. For example, they can test the message is correctly written or that the log level is correctly set depending on the function called.

sduenas (changes requested):
In overall the PR looks good to me but it needs some changes. For example, I think all `log` messages should be set to `debug` (I added some comments about it). Also, please rebase the PR because is failing due `flake8`checks that we already fixed in a previous commit.

evamillan:
I updated the PR adding tests and replacing the `log` messages with `debug`.

sduenas (approved):
LGTM

Issue #462, opened 2021-02-24
Poetry support


sduenas:
We're replacing the way packages are generated in GrimoireLab. The new method will be using poetry. The goal of this task is to add support to it.

sduenas:
Fixed by #483 and #485.

PR #485, opened 2021-03-16
[gha] Use poetry for the CI workflow

vchrombie:
This PR updates the workflow configuration to use poetry for the flake8 check and running the test suite whenever the action is triggered.

It also adds flake8 and httpretty as dev dependencies to the pyproject.toml file as these are needed for the workflow.

#462


sduenas (approved):
LGTM

PR #492, opened 2021-03-18
Update README.md

vchrombie:
The current documentation is of the master branch and doesn't match with the latest version.

This PR updates the README.md with the installation steps and how to run the tests for the latest development branch (muggle) for the sortinghat tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,433 +1,95 @@
-# Sorting Hat
+# Sorting Hat [![build](https://github.com/chaoss/grimoirelab-sortinghat/actions/workflows/ci.yml/badge.svg?branch=muggle)](https://github.com/chaoss/grimoirelab-sortinghat/actions/workflows/ci.yml?query=workflow:build+branch:muggle+event:push)
 
 **Warning**: This is the developing branch of the new version of Sorting Hat. We're moving Sorting Hat into a service.
 The documentation below would be totally incorrect or inaccurate. 
-
 
 ## Description
 
 A tool to manage identities.
 
-Sorting Hat maintains an SQL database with identities coming (potentially) from different sources. Identities corresponding to the same real person can be merged in the same unique identity, with a unique uuid. For each unique identity, a profile can be defined, with the name and other data to show for the corresponding person by default.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,418 +15,66 @@ Sorting Hat is a part of the [GrimoireLab toolset](https://grimoirelab.github.io
...
-and `gender_acc`.
-
-The next MySQL statements should be run to update the schema
+We use [Poetry](https://python-poetry.org/docs/) for managing the project.
 
+Install the required dependencies, we recommend to set up a Python virtual environment to run it
 ```
-mysql> ALTER TABLE profiles ADD COLUMN gender VARCHAR(32) DEFAULT NULL
-mysql> ALTER TABLE profiles ADD COLUMN gender_acc INT(11) DEFAULT NULL
+$ pyenv virtualenv 3.6.13 sh_venv
```
sduenas, line 34: I think that instead of running `pyenv`, users can run `poetry install` directly and then run `poetry shell`. If I'm right, `poetry install` creates a python environment for the project that will be activated running `poetry shell`.

README.md
```diff
@@ -16,418 +15,66 @@ Sorting Hat is a part of the [GrimoireLab toolset](https://grimoirelab.github.io
...
-1. Remove the database and/or create a new one with `sortinghat init`
-1. Load data, this will regenerate the UUIDs
-```
-$ sortinghat load orgs.json
-$ sortinghat load identities.json
+(sh_venv)$ ./manage.py runserver --settings=config.settings.devel
 ```
 
-## Basic commands
+#### frontend
```
sduenas, line 51: Start titles with capital letters. Same in `backend`.

sduenas (changes requested):
I think we should add a section with the basic requirements (Python, Django, and Django graphql) and pointing at the poetry file to check the rest of requirements. 

There are some references in the file to `unique identities` too. This term or concept is no longer used. Please replace it by `individuals`.

Also, please check my minor comments.

README.md
vchrombie: Yeah, what you have said is correct. But, I was considering using only python 3.6 for now since we have only this in the CI.

But, I think you are right. I will update the PR.

sduenas (approved):
LGTM

Issue #496, opened 2021-03-24
GraphQL error: 'str' object has no attribute 'decode' 


sduenas:
This error is raised by the mutation `tokenAuth`. For example, after running the next mutation:
```
mutation tokenAuth {
  tokenAuth(username:"root", password:"*****") {
  	token
  }
}
```

I get the next error:
```
{
  "errors": [
    {
      "message": "'str' object has no attribute 'decode'",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "tokenAuth"
      ],
      "extensions": {
        "code": 128
      }
    }
  ],
  "data": {
    "tokenAuth": null
  }
}
```

sduenas:
After some debugging, I was able to catch the exception:

```
Traceback (most recent call last):
  File "/home/sduenas/.cache/pypoetry/virtualenvs/sortinghat-SE9DHfbA-py3.8/lib/python3.8/site-packages/graphql/execution/executor.py", line 452, in resolve_or_error
    return executor.execute(resolve_fn, source, info, **args)
  File "/home/sduenas/.cache/pypoetry/virtualenvs/sortinghat-SE9DHfbA-py3.8/lib/python3.8/site-packages/graphql/execution/executors/sync.py", line 16, in execute
    return fn(*args, **kwargs)
  File "/home/sduenas/.cache/pypoetry/virtualenvs/sortinghat-SE9DHfbA-py3.8/lib/python3.8/site-packages/graphql_jwt/middleware.py", line 81, in resolve
    return next(root, info, **kwargs)
  File "/home/sduenas/.cache/pypoetry/virtualenvs/sortinghat-SE9DHfbA-py3.8/lib/python3.8/site-packages/promise/promise.py", line 759, in wrapper
    return cls(executor)
  File "/home/sduenas/.cache/pypoetry/virtualenvs/sortinghat-SE9DHfbA-py3.8/lib/python3.8/site-packages/promise/promise.py", line 146, in __init__
    self._resolve_from_executor(executor)
  File "/home/sduenas/.cache/pypoetry/virtualenvs/sortinghat-SE9DHfbA-py3.8/lib/python3.8/site-packages/promise/promise.py", line 493, in _resolve_from_executor
    raise e
[...]

PR #497, opened 2021-03-24
Pin PyJWT to 1.7.1 to fix incompatibility problem with django-graphql-jwt

sduenas:
The package "django-graphql-jwt" is incomatible with the latest version of PyJWT (2.0.1). This raises an error when the mutation "totenAuth" is called because django-graphql-jwt is calling PyJWT API incorrectly. This patch fixes the problem reported in #496.

Without this version the next mutation raised an error:

```
mutation tokenAuth {
  tokenAuth(username:"root", password:"*******") {
  	token
  }
}
```
```

  "errors": [
    {
      "message": "'str' object has no attribute 'decode'",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "tokenAuth"
      ],
      "extensions": {
        "code": 128
      }
    }
  ],
  "data": {
    "tokenAuth": null
  }
}

```

Now it returns something like:

```
{
  "data": {
    "tokenAuth": {
      "token": "************************"
    }
  }
}
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -341,17 +341,16 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "pyjwt"
-version = "2.0.1"
+version = "1.7.1"
 description = "JSON Web Token implementation in Python"
 category = "main"
 optional = false
-python-versions = ">=3.6"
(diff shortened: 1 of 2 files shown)


vchrombie (approved):
LGTM :+1:

PR #499, opened 2021-03-29
Fix integrity issues

sduenas:
This PR fixes some integrity issues that we have in the code. First, it fixes a bug which causes to overwrite log objects if the same id is used. Second, it adds a unit test to check if enrollments  with the same data can be inserted twice.


vchrombie:
LGTM :+1:

PR #506, opened 2021-04-05
[jobs] Catch StopIteration for Python 3.7 compatibility

sduenas:
Since Python 3.7 was introduced, `StopIteration` exceptions are raised directly when `next()` doesn't generate a new item. This
patch adds compatibility other Python versions catching this exception.


vchrombie (approved):
LGTM! 
Thanks for the PR @sduenas.

PR #511, opened 2021-04-06
Order individuals by createdAt field

evamillan:
Allows the individuals to be sorted by their creation date, both from the API and from the UI.


sduenas (approved):
LGTM

PR #512, opened 2021-04-06
[parsing/gitdm] Update VALID_LINE_REGEX expression

zhquan:
This change is needed to execute `gitdm2sh`:

- Skip the lines when there is no email.
- Accept one character as organization name.

Tests updated accordingly.


sduenas (approved):
LGTM

PR #516, opened 2021-04-07
Order individuals by profile name

### LINKED ISSUE #514: Order individuals by profile name
The API and the UI should allow users to order individuals by their profile name. This should be sort in ascendant and descendant orders.

evamillan:
Allows individuals to be sorted by their name, both from the API and from the UI.
Fixes #514.


sduenas (approved):
LGTM

PR #518, opened 2021-04-09
[ui] Set enrollment dates when a new affiliation is created

### LINKED ISSUE #515: Set enrollment dates when a new affiliation is created
Users should be allowed to set the dates of a new enrollment when it is created from the UI. The current behavior sets the dates automatically to their default values (`MAX_DEFAULT_DATE` and `MIN_DEFAULT_DATE`).

evamillan:
Creates a `DateInput` component, which displays a text input that opens a date picker when touched, and adds an optional form to set the dates on the enrollment dialogs in `IndividualsTable`, `OrganizationsTable` and `Workspace`. The styles of the dialogs are also unified.
Fixes #515.

### DIFF
--- a/ui/src/components/DateInput.stories.js
+++ b/ui/src/components/DateInput.stories.js
@@ -1,0 +1,59 @@
+import DateInput from "./DateInput.vue";
+
+export default {
+  title: "DateInput",
+  excludeStories: /.*Data$/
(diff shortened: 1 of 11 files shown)


sduenas (approved):
LGTM

PR #522, opened 2021-04-13
Filter individuals by enrollment 

evamillan:
Adds the `enrollment` filter to the `individuals` query, which returns a list of individuals enrolled in the given organization.
The search syntax in the UI is `enrollment:"organization name"`.

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -262,6 +262,7 @@ class IdentityFilterType(graphene.InputObjectType):
     gender = graphene.String(required=False)
     country = graphene.String(required=False)
     source = graphene.String(required=False)
+    enrollment = graphene.String(required=False)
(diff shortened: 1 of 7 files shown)


sduenas (changes requested):
In overall, it looks good. Although, in the tests, would you mind to add two more enrollments for that individual, one with any given name an another with `Bit Company`, so you can check two more things:

-  The filter "bit" returns two enrollments.
- The filter `Bit Company` returns only one and that the case "Organizations that include spaces should be wrapped between double quotes" works.

evamillan:
I updated the PR adding those test cases.

sduenas (approved):
LGTM. Would you mind to rebase it?

evamillan:
Done.

sduenas (approved):
LGTM

PR #529, opened 2021-04-19
Filter individuals by whether or not they have any enrollments

evamillan:
This PR adds the `isEnrolled` filter to the `individuals` query, which returns a list of individuals that are affiliated or not to any organizations.


sduenas (approved):
LGTM

PR #530, opened 2021-04-21
Add documentation for GraphQL API

evamillan:
Adds descriptions to the GraphQL queries and mutations.


sduenas (changes requested):
In general it looks good. There are some periods that are  not at the end of the line and I think you should add some examples about dates and country codes.

evamillan:
Thanks for the comments. I've added the periods and examples.

sduenas (approved):
LGTM

PR #543, opened 2021-06-10
[utils/create_fixture] Create fixture JSON file of a database

### LINKED ISSUE #542: Script to migrate data from version 0.7.x to new version
Due to data model has changed from one version to another and SQLAlchemy is not used anymore, we need to create a script that allows to export data and import it in the new sortinghat.

We think the best solution is to create a fixture file. This file can be imported directly using django commands. This would be even faster than creating export/import methods or to call mutations directly.

zhquan:
This script creates a fixture JSON file given the old SortingHat (0.7.x)
database to load into the new SortingHat using Django.

Execute:
```
$ python3 create_fixture.py localhost test_sh
[2021-06-10 17:29:11,461][INFO] Start creating fixture file for test_sh
[2021-06-10 17:29:21,252][INFO] Fixture file created: test_sh_fixture.json
[2021-06-10 17:29:21,252][INFO] Total duration: test_sh_fixture.json
```

First, you have to create a new database.
```
mysql> CREATE DATABASE new_sh CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci;
```

Load fixture JSON file using Django
```
$ python3 manage.py loaddata test_sh_fixture.json --settings=config.settings.devel
Installed 148542 object(s) from 1 fixture(s)
```

Fixes #542

### DIFF
--- a/utils/create_sh_0_7_fixture.py
+++ b/utils/create_sh_0_7_fixture.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 Bitergia
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+# Authors:
+#   Quan Zhou <quan@bitergia.com>
+#
+
+
+import argparse
+from datetime import datetime, timezone
+import json
+import logging
+import sys
+
+import MySQLdb
+
+
+DESC_MSG = """This script creates a fixture JSON file given the old SortingHat (0.7.x)
+database to load into the new SortingHat using Django.
+
+    Execute:
+    ```
+    $ python3 create_sh_0_7_fixture.py test_sh -o test_sh_fixture.json
+    [2021-06-10 17:29:11,461][INFO] Start creating fixture file for test_sh
+    [2021-06-10 17:29:21,252][INFO] Fixture file created to test_sh_fixture.json
+    ```
+
+    First you have to create the new database.
+    ```
+    mysql> CREATE DATABASE new_sh CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci;
+    ```
+    Execute Django migrate.
+    ```
+    $ python3 manage.py migrate --settings=config.settings.devel
+    ```
+    Load fixture JSON file using Django
+    ```
+    $ python3 manage.py loaddata test_sh_fixture.json --settings=config.settings.devel
+    Installed 148542 object(s) from 1 fixture(s)
+    ```
+"""
+
+TIMESTAMP_FORMAT = '%Y-%m-%d %H:%M:%S%z'
+
+# Mappings
+
+COUNTRY_MAPPING = {
+    "code": "code",
+    "name": "name",
+    "alpha3": "alpha3"
+}
+
+DOMAIN_MAPPING = {
+    "id": "id",
+    "domain": "domain",
+    "is_top_domain": "is_top_domain",
+    "organization_id": "organization"
+}
+
+ENROLLMENT_MAPPING = {
+    "id": "id",
+    "start": "start",
+    "end": "end",
+    "organization_id": "organization",
+    "uuid": "individual"
+}
+
+IDENTITY_MAPPING = {
+    "id": "uuid",
+    "name": "name",
+    "email": "email",
+    "username": "username",
+    "source": "source",
+    "uuid": "individual",
+    "last_modified": "last_modified"
+}
+
+MATCHING_BLACKLIST_MAPPING = {
+    "excluded": "excluded"
+}
+
+ORGANIZATION_MAPPING = {
+    "id": "id",
+    "name": "name"
+}
+
+PROFILE_MAPPING = {
+    "uuid": "individual",
+    "name": "name",
+    "email": "email",
+    "gender": "gender",
+    "gender_acc": "gender_acc",
+    "is_bot": "is_bot",
+    "country_code": "country"
+}
+
+INDIVIDUAL_MAPPING = {
+    "uuid": "mk",
+    "last_modified": "last_modified"
+}
+
+TABLE_MAPPING = {
+    "countries": {
+        'model': "core.country",
+        'mapping': COUNTRY_MAPPING
+    },
+    "domains_organizations": {
+        'model': "core.domain",
+        'mapping': DOMAIN_MAPPING
+    },
+    "enrollments": {
+        'model': "core.enrollment",
+        'mapping': ENROLLMENT_MAPPING
+    },
+    "identities": {
+        'model': "core.identity",
+        'mapping': IDENTITY_MAPPING
+    },
+    "matching_blacklist": {
+        'model': "core.matchingblacklist",
+        'mapping': MATCHING_BLACKLIST_MAPPING
+    },
+    "organizations": {
+        'model': "core.organization",
+        'mapping': ORGANIZATION_MAPPING
+    },
+    "profiles": {
+        'model': "core.profile",
+        'mapping': PROFILE_MAPPING
+    },
+    "uidentities": {
+        'model': "core.individual",
+        'mapping': INDIVIDUAL_MAPPING
+    }
+}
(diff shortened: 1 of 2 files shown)


utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+        model = "core.matchingblacklist"
+    elif table == "organizations":
+        model = "core.organization"
+    elif table == "profiles":
+        model = "core.profile"
+    elif table == "uidentities":
+        model = "core.individual"
+    else:
+        model = None
+        logger.error("[ERROR] Table name unknown {}".format(table))
```
sduenas, line 208: Instead of doing this, create a dictionary `MODEL_MAPPING` like:

```
MODEL_MAPPING = {
    'countries' : 'core.country',
    'domains_organizations': 'core.domain'
    ...
}
```
Then you can do a `try..except` to catch the error if the table is unknown.

utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+        elif h == "uuid" and table != "uidentities":
+            field = "individual"
+        elif h == "uuid" and table == "uidentities":
+            field = "mk"
+        elif h == "organization_id":
+            field = "organization"
+        elif h == "country_code":
+            field = "country"
+        else:
+            field = h
```
sduenas, line 257: You can also create a mapping table here to the kind of the same.

utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+    fixtures = []
+    for table in tables:
+        logger.debug("Table {} start".format(table))
+        fixtures += get_fixture(db, time_str, table)
+        logger.debug("Table {} finished".format(table))
+    db.close()
+
+    file_name = database + "_fixture.json"
+    write_json(fixtures, file_name)
+
```
sduenas, line 93: You can move all of this code to a function `generate_sortinghat_fixture`, so it's easy to read and the main function only contains simple code.

utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+    ```
+    $ python3 manage.py loaddata test_sh_fixture.json --settings=config.settings.devel
+    Installed 148542 object(s) from 1 fixture(s)
+    ```
+    """
+    args = parse_args()
+    host = args.host
+    user = args.user
+    password = args.password
+    database = args.database
```
sduenas, line 69: You can ignore these variables an use `args.*` directly because you are passing them just once to create the connection.

utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+import argparse
+from datetime import datetime, timezone
+import json
+import logging
+
+import pymysql
+
+QUERY_DESC = 'desc {};'
+QUERY_SEARCH = 'SELECT * FROM {};'
+QUERY_TABLES_SHOW = 'show tables;'
```
sduenas, line 33: These constants can be moved to the functions that use them because they are not reused by other functions.

utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+#   Quan Zhou <quan@bitergia.com>
+#
+
+
+import argparse
+from datetime import datetime, timezone
+import json
+import logging
+
+import pymysql
```
sduenas, line 29: Instead of `pymysql` we should use `mysqlclient`, which is already a sortinghat dependency.

utils/create_fixture.py
```diff
@@ -0,0 +1,271 @@
...
+        count = 0
+        for d in data:
+            new = {
+                "fields": {},
+                "model": model
+            }
+            for index, h in enumerate(headers):
+                value = d[index]
+                if type(d[index]) == datetime:
+                    value = d[index].strftime(STRFTIME)
```
sduenas, line 170: I think data can be exported into a dictionary that for each row, contains the name of the columns and their values. Check [DictCursor](http://mysql-python.sourceforge.net/MySQLdb-1.2.2/public/MySQLdb.cursors.DictCursor-class.html)

sduenas (changes requested):
Please check my review. I think there are ways to improve the code and make it easier to read.

utils/create_fixture.py
zhquan: `fetchallDict(self): Fetch all available rows as a list of dictionaries. Deprecated: Use `fetchall()` instead. Will be removed in 1.3.`

Also, we have to change the value when the type is datetime

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+    mysql> CREATE DATABASE new_sh CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci;
+    ```
+    Execute Django migrate.
+    ```
+    $ python3 manage.py migrate --settings=config.settings.devel
+    ```
+    Load fixture JSON file using Django
+    ```
+    $ python3 manage.py loaddata test_sh_fixture.json --settings=config.settings.devel
+    Installed 148542 object(s) from 1 fixture(s)
```
sduenas, line 155: Maybe this text can go in the output of `--help`?

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+    parser.add_argument('-g', dest='debug_mode', action='store_true',
+                        default=False,
+                        help="Enable debug mode")
+    args = parser.parse_args()
+
+    return args
+
+
+def generate_sortinghat_fixtures(conn):
+    """Generate SortingHat fixture
```
sduenas, line 200: Add a period at the end of the sentence.

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+
+    :param conn: MySQL connection
+
+    :return fixture
```
sduenas, line 204: This should say that it returns a list of dict objects that contain the fixture data.

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+    for table in TABLE_MAPPING.keys():
+        logger.debug(f"Table {table} start")
+        fixtures += generate_fixture(conn, table, timestamp)
+        logger.debug(f"Table {table} finished")
+
+    return fixtures
+
+
+def generate_fixture(conn, table, timestamp):
+    """Create fixture given a table
```
sduenas, line 219:
```suggestion
    """Create fixture data of the given table.
```

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+
+    :param conn: MySQL connection
+    :param table: table
+    :param timestamp: Datetime now
+
+    :return: fixture
```
sduenas, line 225: Same comment as before.

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+    ]
+    logger.debug(f"{table} total items added {len(fixtures)}")
+
+    return fixtures
+
+
+def fetch_data(conn, table):
+    """Fetch data from MySQL and return a iterator with the next format:
+    - Keys: column names.
+    - Values: row values.
```
sduenas, line 239:
```suggestion
    """Fetch data from MySQL and return a iterator with the next format:
    
      - Keys: column names.
      - Values: row values.
```

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+        "key1": "value1",
+        "key2": "value2",
+        ...
+    }
+    ```
+
+    :param conn: MySQL connection
+    :param table: table name
+
+    :return: iterator
```
sduenas, line 251:
```suggestion
    :return: iterator of table rows
```

utils/create_sh_0_7_fixture.py
```diff
@@ -0,0 +1,323 @@
...
+            yield data
+    except Exception as ex:
+        logger.error(ex)
+        raise ex
+
+
+def map_row_data(data, table, timestamp):
+    """Map row data to a Django fixture format
+    ```
+    {
```
sduenas, line 273:
```suggestion
    """Map row data to the Django fixture format.
    
    The format is:
    
```
    {
```

File: utils/create_sh_0_7_fixture.py [hunk 16]
```diff
@@ -0,0 +1,323 @@
...
+    entry = {
+        "fields": fields,
+        "model": TABLE_MAPPING[table]['model']
+    }
+
+    return entry
+
+
+def write_json(fd, data):
+    try:
```
sduenas, line 306:
```suggestion
def write_json(fd, data):
    """Write data in JSON format to a file descriptor."""
    
    try:
```

sduenas (changes requested):
The PR looks ok but it needs minor changes. Also, we need to add an entry in the README file called `Compatibility between versions` explaining how to migrate the information of 0.7 to newer versions. Take as reference the entry in `master` branch.

sduenas (approved):
LGTM

sduenas:
@zhquan do you mind to rebase it so I can merge it?

PR #549, opened 2021-06-18
[ui] Fix synchronous request warning

### LINKED ISSUE #531: Deprecation warning about a synchronous request
There is a warning about a synchronous request being made in `ui/src/main.js` line 17.
 According to the warning, it says, I quote,  `[Deprecation] Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience.`
The solution would be simply to change our current request to `async.` However, since the request was made `sync` from the beginning, I would assume there is a purpose. In any case, I have my doubts on this so I created this issue.

evamillan:
Fixes #531, replacing the deprecated synchronous `XMLHttpRequest` to get the CSRF token with the newer `fetch` API. The request is asynchronous, but the token is retrieved after it is finished, using `then()`.

### DIFF
--- a/ui/src/main.js
+++ b/ui/src/main.js
@@ -13,58 +13,56 @@ import { ApolloLink } from "apollo-link";
 import Logger from "./plugins/logger";
 
 // Force HTTP GET to the Django Server for getting the csrf token
-let xmlHttp = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


sduenas (approved):
LGMT

PR #550, opened 2021-06-18
[gha] Update tests workflow

vchrombie:
This PR renames the ci github actions workflow to avoid ambiguity. It also adds a rule to restrict the tests workflow to run when a release tag is pushed to the repository to avoid redundant workflows.

It also adds additional python versions 3.7, 3.8 for running the test suite.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,9 +1,11 @@
-name: build
+name: tests
 
-on: 
+on:
   push:
     branches:
       - muggle
+    tags:
+      - '!*.*.*'
   pull_request:
     branches:
       - muggle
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -1,9 +1,11 @@
-name: build
+name: tests
 
-on: 
+on:
   push:
     branches:
       - muggle
```
vchrombie, line 6: when the `muggle` is merged into `master`, this can be replaced with 

```
     branches:
       - '**'
```

sduenas (approved):
LGTM, would you rebase it?

vchrombie:
> LGTM, would you rebase it?

Yup, I will do it right now.

Issue #548, opened 2021-06-17
Create a Groups model to handle organizational hierarchies


Rashmi-K-A:
As discussed in #541, a new model called `Group` needs to be created to store subdivisions/departments/groups of an organization. 

- [x]  Create model `Group`
- [x]  Implement `add_group`, `find_group`, `delete_group` methods in `db.py`
- [x]  Implement `add_group`, `find_group`, `delete_group` methods in `api.py`

sduenas:
Fixed by #552 .

Issue #541, opened 2021-05-25
Evaluate how to store Organization information


Rashmi-K-A:
- [x] Understand existing Organization model
- [x] Understand existing problems with the model and gaps that need to be addressed
- [ ] Evaluate how to store organizational hierarchies 
- [ ] Evaluate how to store organization aliases

Rashmi-K-A:
The proposal here modifies the existing Organization model to handle hierarchical data.

An alternate approach is to keep the Organization model as it is and create a new model for groups/divisions/departments. We need to evaluate how to proceed with this alternate approach and leave the existing model untouched.

Rashmi-K-A:
New approach:
  1. Create a new `Group` model containing just `name` field for now. 
  2. Update `Organization` model to have a foreign key to the `Group` model.
  3.  Whenever a new organization is created, a new root is also created in the table for storing groups. 

Since we are linking groups and organizations, we need to update the existing methods in the code that use the organizations table in the DB to use the treebeard API as well. 

The list of such methods that need to be updated are:
`add_organization` -  On organization creation, we will also create a new root node in groups table.
`delete_organization` - On organization deletion, we will delete the groups tree for the organization.


POC with updated database schema:
`groups_orgs` is the sample table I created that would hold group hierarchy information as per treebeard. The only fields that we have added are `name` and the ones inherited from EntityBase model
<img width="1252" alt="Screenshot 2021-06-08 at 9 00 06 PM" src="https://user-images.githubusercontent.com/39820442/121214265-8fec1800-c89c-11eb-89bd-e335927cc7a0.png">


@mafesan @vchrombie @sduenas @evamillan  any thoughts on these?

Rashmi-K-A:
In terms of naming, what should we call this new Model? If we were to call it Group, the only thing we need to remember is that we cannot name the database table as `group` or `groups`. However, to name the model, I think Group should do fine.

sduenas:
> In terms of naming, what should we call this new Model? If we were to call it Group, the only thing we need to remember is that we cannot name the database table as `group` or `groups`. However, to name the model, I think Group should do fine.

What about `team` and `teams`?

sduenas:
> 
>     3. Whenever a new organization is created, a new root is also created in the table for storing groups.

What's the reason for this? Why does it need a root entry?

Rashmi-K-A:
> > ```
> > 3. Whenever a new organization is created, a new root is also created in the table for storing groups.
[...]

PR #552, opened 2021-06-22
Add Teams model 

Rashmi-K-A:
This PR is in reference to issue #548 for creating a ~Groups~ **Teams** model to handle organizational hierarchies. This PR includes changes to the following files:

- models.py - add Team model
- db.py - add internal methods `add_team`, `find_team` and `delete_team`
- api.py - add methods `add_team`, `delete_team`

#541  has more information on the discussion around how to handle organizational hierarchies.


sortinghat/core/migrations/0009_auto_20210622_1033.py
```diff
@@ -0,0 +1,37 @@
+# Generated by Django 2.2 on 2021-06-22 15:33
```
sduenas: We started using Django 3.1. Please squash the commit you have later with this one.

sortinghat/core/models.py
```diff
@@ -143,6 +145,18 @@ def __str__(self):
         return self.name
 
 
+class Group(MP_Node, EntityBase):
```
sduenas, line 148: Thought we were going to name this class `OrganizationGroup` or something similar. Although if it's no mandatory to assing it to an organization, probably it doesn't make sense. Then, we should look for a different name for the table.

sortinghat/core/models.py
```diff
@@ -143,6 +145,18 @@ def __str__(self):
...
+    name = CharField(max_length=MAX_SIZE_CHAR_INDEX)
+    organization = ForeignKey(Organization, related_name='groups', on_delete=CASCADE, blank=True, null=True)
```
sduenas, line 150: It would be better to split this line in:

```
organization = ForeignKey(Organization, related_name='groups', on_delete=CASCADE,
                          blank=True, null=True)
```

tests/test_model.py
```diff
@@ -106,6 +107,61 @@ def test_last_modified(self):
...
+          org = Organization.objects.create(name='Example')
+          group = Group.add_root(name='subGroup1', organization=org)
+          group.add_child(name='subGroup1', organization=org)
+
+
+    def test_null_organizations(self):
+        """Check if groups can be created without organizations"""
+
+        group = Group.add_root(name='subGroup1')
+        self.assertIsInstance(group,Group)
```
sduenas, line 126: Have you tried to create a second instance with the same name and the organization as null? I think it will allow you when that shouldn't be the case.

sortinghat/core/db.py
```diff
@@ -290,6 +291,77 @@ def delete_organization(trxl, organization):
                        target=op_args['organization'])
 
 
+def add_group(trxl, organization, parent, group_name):
```
sduenas, line 294: As `organization` and `parent` can be `None`, I think it's better to reformulate the definition of the function this way:
```
def add_group(trxl, group_name, organization=None, parent=None):
```

sortinghat/core/db.py
```diff
@@ -290,6 +291,77 @@ def delete_organization(trxl, organization):
...
+    except django.db.utils.IntegrityError as exc:
+        _handle_integrity_error(Group, exc)
+
+    trxl.log_operation(op_type=Operation.OpType.ADD, entity_type='group',
+                       timestamp=datetime_utcnow(), args=op_args,
+                       target=op_args['group_name'])
+    return group
+
+
+def find_group(organization, group_name):
```
sduenas, line 338: Same as before. It should be:

```
def find_group(group_name, organization=None):
```

sortinghat/core/db.py
```diff
@@ -290,6 +291,77 @@ def delete_organization(trxl, organization):
...
+    """Finds a group in an organization, if organization is passed. Otherwise,
+    searches for group with name=`group_name` and organization=`null`.
+
+    Finds a group by its name and organization in the database. When the
+    group does not exist, the function will raise a `NotFoundError`.
```
sduenas, line 343: This also should be:

""Find a group in an organization.

Finds a group by its name in the database by its name. If organization is passed, it looks for a group in that organization. When the group does not exist, the function will raise a `NotFoundError`.

sortinghat/core/db.py
sduenas: This function is should go also where the other `find_*` functions are.

tests/test_db.py
```diff
@@ -203,6 +209,59 @@ def test_organization_not_found(self):
...
+
+        self.assertIsInstance(group, Group)
+        self.assertEqual(group.organization.name, 'Example')
+        self.assertEqual(group.name, 'Example Subgroup')
+        self.assertEqual(group.get_parent(), None)
+
+    def test_group_not_found(self):
+        """Check if an error is raise if group is not found"""
+        with self.assertRaisesRegex(NotFoundError, GROUP_NOT_FOUND_ERROR):
+            db.find_group(self.org, 'subGroup')
```
sduenas, line 232: Leave a blank line between the docstring and the code.

tests/test_db.py
```diff
@@ -203,6 +209,59 @@ def test_organization_not_found(self):
...
+
+    def test_org_name_none(self):
+        """Check if groups linked to no organizations can be found"""
```
sduenas, line 235: Same as before.

tests/test_db.py
```diff
@@ -578,6 +637,198 @@ def test_operations(self):
...
+
+    def test_add_subgroup(self):
+        """Check if a new subgroup is added for specified group"""
+
+        group_name = 'subgroup'
+        group = db.add_group(self.trxl, self.org, None, group_name)
+        subgroup = db.add_group(self.trxl, self.org, group, 'childgroup')
+        self.assertIsInstance(subgroup, Group)
+        self.assertEqual(subgroup.name, 'childgroup')
+        self.assertEqual(subgroup.organization, self.org)
```
sduenas, line 670: Check also that the parent is `subgroup`.

tests/test_db.py
```diff
@@ -578,6 +637,198 @@ def test_operations(self):
...
+        """Load initial dataset"""
+
+        self.user = get_user_model().objects.create(username='test')
+        self.ctx = SortingHatContext(self.user)
+
+        self.trxl = TransactionsLog.open('delete_group', self.ctx)
+        self.org = Organization.objects.create(name='Example')
+
+    def test_delete_group(self):
+        """Check whether it deletes a group"""
```
sduenas, line 791: Leave a blank line between docstring and code.

sortinghat/core/api.py
```diff
@@ -565,6 +568,73 @@ def add_domain(ctx, organization, domain_name, is_top_domain=True):
     return domain
 
 
+@django.db.transaction.atomic
+def add_group(ctx, organization, parent_name, group_name):
```
sduenas, line 572: Because `organization` and `parent_name` can be None, reorganize the definition as explained before.

sortinghat/core/api.py
```diff
@@ -644,6 +714,56 @@ def delete_domain(ctx, domain_name):
     return domain
 
 
+@django.db.transaction.atomic
+def delete_group(ctx, organization, group_name):
```
sduenas, line 718: Same comment with the definition of the function as before.

sortinghat/core/api.py
```diff
@@ -644,6 +714,56 @@ def delete_domain(ctx, domain_name):
...
+    """Remove a group from the registry.
+
+    This function removes the given group of an organization from the registry.
+    Deleting a group also deletes all its subgroups in the organization.
+    Both Organization and Group must exist in the registry. Otherwise, function
+    will raise a `NotFoundError` exception.
```
sduenas, line 724: You should reword this because the group doesn't need to belong to an organization.

tests/test_api.py
```diff
@@ -2026,6 +2029,155 @@ def test_operations(self):
...
+
+    def test_organization_is_none(self):
+        """Check if it fails when organization name is `None`"""
+
+        api.add_group(self.ctx, "Example", None, "suborg")
+        trx_date = datetime_utcnow()
+        with self.assertRaisesRegex(InvalidValueError, GROUP_ORG_NAME_MISSING):
+            api.delete_group(self.ctx, None, "suborg")
+        transactions = Transaction.objects.filter(created_at__gt=trx_date)
+        self.assertEqual(len(transactions), 0)
```
sduenas, line 2133: Please, organize this test (and the other ones) so we can read them better. For example:

```
api.add_group(self.ctx, "Example", None, "suborg")

trx_date = datetime_utcnow()

with self.assertRaisesRegex(InvalidValueError, GROUP_ORG_NAME_MISSING):
     api.delete_group(self.ctx, None, "suborg")

transactions = Transaction.objects.filter(created_at__gt=trx_date)
self.assertEqual(len(transactions), 0)
```

sduenas (changes requested):
In general it looks good but some changes are needed. Please also check and modify the style in the tests and with the docstrings.

sortinghat/core/models.py
Rashmi-K-A: What do you suggest? OrganizationGroup doesnt make much sense since we discussed that some group don't have to be linked to organizations.

sortinghat/core/models.py
Rashmi-K-A: Got it, I forgot to format the file and break down lines with >80 chars, will do

sortinghat/core/models.py
sduenas: To be honest, I'm more towards `Team` to be honest. Many people know about GitHub teams so it's easy to understand what it is.

Rashmi-K-A: Works for me! I had an initial impression that Team might only feel like small groups (I don't know much about Github teams). However, if you think it will work, we can make the model into Team and call the table 'teams'

tests/test_model.py
Rashmi-K-A: yes, just caught this. I should have put in a unit test for this one. I think I can override the models validate_unique or clean function to catch this.  What are your thoughts?

Rashmi-K-A: @sduenas I have added a `validate_unique` function to the model and called `group.full_clean()` before saving the objects in `add_group` function in `db.py`. It catches this error. Do you think that will suffice? I will add sufficient tests as well.

tests/test_api.py
Rashmi-K-A: Done 👍

sortinghat/core/api.py
Rashmi-K-A: Done 👍

sortinghat/core/api.py
Rashmi-K-A: Done 👍

sortinghat/core/api.py
Rashmi-K-A: Done 👍

tests/test_db.py
Rashmi-K-A: Done 👍

sortinghat/core/db.py
Rashmi-K-A: Done 👍

sortinghat/core/db.py
Rashmi-K-A: Done 👍

tests/test_db.py
Rashmi-K-A: Done 👍

tests/test_db.py
Rashmi-K-A: Done 👍

tests/test_db.py
Rashmi-K-A: Done 👍

sortinghat/core/db.py
Rashmi-K-A: Done 👍

sortinghat/core/models.py
Rashmi-K-A: Done 👍

tests/test_model.py
sduenas: We need a test for that case to check that it doesn't let it create two groups/teams with the same name and org = NULL.

Rashmi-K-A: Done :+1

sortinghat/core/db.py
```diff
@@ -822,13 +924,13 @@ def _handle_integrity_error(model, exc):
...
     m = re.match(_MYSQL_DUPLICATE_ENTRY_ERROR_REGEX,
                  exc.__cause__.args[1])
     if not m:
         raise exc
 
     entity = model.__name__
     eid = m.group('value')
-
+    if model == Team:
+        eid = m.group('value').split('-')[0]
```
sduenas, line 935:
```suggestion

    if model == Team:
        eid = m.group('value').split('-')[0]
        
```

sortinghat/core/models.py
```diff
@@ -143,6 +147,25 @@ def __str__(self):
...
+
+    class Meta:
+        db_table = 'teams'
+        unique_together = ('name', 'organization',)
+
+    def validate_unique(self, exclude=None):
+        if not self.organization:
+            objects = Team.objects.all().filter(name=self.name, organization=self.organization)
+            if len(objects) > 0:
+                raise ValidationError(f"Team {self.name} already exists in the registry")
```
sduenas, line 163: This function is not really supposed to do that. Do you think we can find something else? This method will slow the process. Usually it's better to try to write and get the error that ask first.

tests/test_api.py
```diff
@@ -2026,6 +2029,177 @@ def test_operations(self):
...
+        trx_date = datetime_utcnow()
+
+        with self.assertRaisesRegex(NotFoundError, ORGANIZATION_NOT_FOUND_ERROR.format(name='Exampe')):
+              api.add_team(self.ctx, "subteam", "Exampe", None, )
+
+        transactions = Transaction.objects.filter(created_at__gt=trx_date)
+        self.assertEqual(len(transactions), 0)
+
+    def test_find_parent(self):
+        """ Check if parent is able to be found"""
```
sduenas, line 2094:
```suggestion
        """ Check if parent is able to be found"""
        
```

tests/test_db.py
```diff
@@ -578,6 +640,226 @@ def test_operations(self):
...
+
+        team1 = db.add_team(self.trxl, "team1", self.org, None)
+        team2 = db.add_team(self.trxl, "team2", self.org, None)
+
+        teams = Team.objects.all()
+
+        self.assertIsInstance(team1, Team)
+        self.assertIsInstance(team2, Team)
+        self.assertEqual(len(teams), 2)
+
```
sduenas, line 779: Remove this line here.

sduenas (changes requested):
In overall it looks ok but I don't like much the way you're solving the integrity problem with the same name team without organization.
You also need to remove that merge commit that is there.

sortinghat/core/models.py
Rashmi-K-A: Yes, we need to figure out some way to add this validation though. Because if teams have orgs as NULL, the inbuilt validation just doesn't work. This is going to test for every node that has orgs as NULL and I agree that its going to affect the performance if we have too many teams without orgs.  Is there any alternative that you can suggest or shall I dig around more and see what I can find?

Rashmi-K-A:
> You also need to remove that merge commit that is there.
Yes, I have a call with Venu today to fix that. I thought its better to let him show me how to fix that the first time!

sortinghat/core/models.py
sduenas: I don't exactly remember how we do this with the identities. The problem should be the same. Can you check it? Because I couldn't find it.

Rashmi-K-A:
@sduenas  I have fixed the commits (thanks to @vchrombie!). We were wondering about the changelog entry. Should that be added as part of this PR?

sduenas:
> @sduenas I have fixed the commits (thanks to @vchrombie!). We were wondering about the changelog entry. Should that be added as part of this PR?

Maybe we can skip that part until the implementation with graphql + ui is done.

sduenas:
> @sduenas I have fixed the commits (thanks to @vchrombie!). We were wondering about the changelog entry. Should that be added as part of this PR?

@Rashmi-K-A would you mind to change the style of the commits again? The body of the message should be limited to 70 characters or so. Now they are very long and only good for large screens.

vchrombie:
@Rashmi-K-A 
Good work with the re-writing the commits, but I think we have some more things to do.
You can truncate the commit description into multiple lines.

vchrombie:
> @Rashmi-K-A would you mind to change the style of the commits again? The body of the message should be limited to 70 characters or so. Now they are very long and only good for large screens.

Jinx! 😄 🥂

vchrombie:
@Rashmi-K-A let me know if you need help with anything.

Rashmi-K-A:
Thanks @vchrombie @sduenas !! I have fixed it now 💯  Let me know if I need to change anything else!

sduenas (approved):
LGTM

sduenas:
Merged. Thanks for your contribution!

PR #554, opened 2021-06-23
Improve error messages

### LINKED ISSUE #551: Better error messages
When there's an error in the UI the error messages usually make reference to GraphQL. We need better error messages because most users won't care about GraphQL neither will know what that means.
We need better messages so users will understand what the problem is and how to fix it.

This is an example:

harry

evamillan:
This PR adds a plugin to improve the error messages displayed on the UI, removing unnecessary text (eg. "GraphQL Error") and replacing some error messages with more readable ones. It also adds error messages to some actions that didn't have them and a custom code for invalid JWT credentials.
Closes #551.


sduenas (approved):
LGTM

PR #555, opened 2021-06-30
Add recommenderexcludedlist

### LINKED ISSUE #553: Recommender Exclusion List
The idea of this list is to exclude from the recommendations process those identities that have any of those values. For example, we know that usernames such as `root` or email addresses like `noreply.github.com` are generic and can create false positives, so they shouldn't be used to recommend matches.

The user should be able to add/remove elements from the list. There should be a query that allows to get this list.

This list will replace the former `blacklist`. We can call it in the model `RecommenderExclusionList`

zhquan:
This code skips matching for all those profiles that one of the values of 'name', 'username', or 'email' is in 'RecommenderExclusionList'.

New query:
- recommenderExclusionList

New mutations:
- addRecommenderExclusionList(exclude: String)
- deleteRecommenderExclusionList(exclude: String)

`exclude` ( by default is `True`) parameter added to:
- unify
- recommendMatches

Fixes #553


sortinghat/core/migrations/0010_auto_20210624_1234.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 3.1 on 2021-06-24 12:34
```
sduenas, line 1: This commit should be squashed with the first one. Also, you should try to change the name of the migration. See https://docs.djangoproject.com/en/3.2/ref/django-admin/#cmdoption-makemigrations-name

sortinghat/core/models.py
```diff
@@ -233,8 +233,12 @@ def __str__(self):
         return '%s - %s' % (self.individual.mk, self.organization.name)
 
 
-class MatchingBlacklist(EntityBase):
-    excluded = CharField(max_length=MAX_SIZE_CHAR_FIELD, primary_key=True)
+class RecommenderExclusionList(EntityBase):
```
sduenas, line 236: I think we should rename this to `RecommenderExclusionTerm` because the object stores the term. It's not a list.

sortinghat/core/models.py
```diff
@@ -233,8 +233,12 @@ def __str__(self):
...
+    excluded = CharField(max_length=MAX_SIZE_CHAR_INDEX)
 
     class Meta:
-        db_table = 'matching_blacklist'
+        db_table = 'recommender_exclusion_list'
```
sduenas, line 240: This should be kept with this name because that's the table where these objects are stored.

sortinghat/core/db.py
```diff
@@ -196,6 +197,44 @@ def find_domain(domain_name):
         return domain
 
 
+def find_recommenderExclusionList(exclusion):
```
sduenas, line 200: I don't think we need this function because add/remove functions should try to do their job even if the term doesn't exist.

sortinghat/core/recommendations/matching.py
```diff
@@ -26,15 +26,15 @@
...
+from ..db import (find_individual_by_uuid, fetch_recommenderExclusionList)
 from ..errors import NotFoundError
 from ..models import Identity
 
 
 logger = logging.getLogger(__name__)
 
 
-def recommend_matches(source_uuids, target_uuids, criteria, verbose=False):
+def recommend_matches(source_uuids, target_uuids, criteria, exclude, verbose=False):
```
sduenas, line 37: `exclude` should be activated by default.

sortinghat/core/recommendations/matching.py
```diff
@@ -149,6 +155,15 @@ def _to_df(data_set):
...
         return df.sort_values(['individual'])
 
+    def _apply_recommenderExclusionList(df):
+        """Apply RecommenderExclusionList to returns the dataframes that do not match
+        `name`, `username`, or `email` with this excluded list"""
+        excluded = fetch_recommenderExclusionList()
+        df_exclude_name = df[~df['name'].isin(excluded)]
+        df_exclude_username = df_exclude_name[~df_exclude_name['username'].isin(excluded)]
+        df_exclude_email = df_exclude_username[~df_exclude_username['email'].isin(excluded)]
+        return df_exclude_email
```
sduenas, line 165: Is not possible to do everything on the same step?

sduenas (changes requested):
In overall the PR looks good, minor changes are needed, but I wonder if it's possible to reduce the complexity. My point is that as this is part of the recommendation system, probably everything should be moved there and we can have an object, for example, `RecommendationExclusionList` that has methods to add/remove/list terms in the database. That will mean, no api/db methods but you can refactorize them on this new object.

This can "break" the way we were developing operations with the database but I think in this case it make sense because, although it's part of the core, it isn't really needed to operate with sortinghat.

sduenas (changes requested):
Additionally to my comments, I miss recommender tests.

sduenas (approved):
LGTM

PR #556, opened 2021-07-02
Gender recommendations

### LINKED ISSUE #471: Gender recommendations
The new SortingHat service should provide recommendations about gender based on individuals name. This is something that is available in previous versions using `autogender` command and `genderize.io` service.

evamillan:
This PR adds a `recommendGender` mutation, which returns a job with a possible gender for each given individual uuid based on their profile name. The recommendations are provided by the genderize.io API.

Example of the mutation:
```
mutation {
  recommendGender(uuids: ["uuid"], apiToken: "token") {
    jobId
  }
}
```
Example of the job query:
```
query {
  job(jobId: "id") {
    jobId
    status
    jobType
    result {
      __typename
      ... on GenderRecommendationType {
          uuid
          gender
	  accuracy
      }
    }
    errors
  }
}
```

The individuals' genders can be autocompleted using these same recommendations with the `autogender` mutation.
Example of the mutation:
```
mutation {
  autogender(uuids:["uuid1", "uuid2"], apiToken: "token") {
    jobId
  }
}
```
Example of the job result query:
```
query {
  job(jobId: "id") {
    status
    jobType
    result {
      ... on AutogenderResultType {
        uuid
        gender
        accuracy
      }
    }
    errors
  }
}
```
Closes #471.

### DIFF
--- a/sortinghat/core/jobs.py
+++ b/sortinghat/core/jobs.py
@@ -220,6 +220,49 @@ def recommend_matches(ctx, source_uuids, target_uuids, criteria, verbose=False):
 
 
 @django_rq.job
+def recommend_gender(ctx, uuids, api_token=None):
+    """Generate a list of gender recommendations from a set of individuals.
+
+    This job generates a list of recommendations with the
+    probable gender of the given individuals.
+
+    :param ctx: context where this job is run
+    :param uuids: list of individuals identifiers
+    :api_token: genderize.io API token
+
+    :returns: a dictionary with the recommended gender and accuracy of the
+        prediction for each individual.
+    """
+    job = rq.get_current_job()
+
+    logger.info(f"Running job {job.id} 'recommend gender'; ...")
+
+    results = {}
+    job_result = {
+        'results': results
+    }
+
+    engine = RecommendationEngine()
+
+    job_ctx = SortingHatContext(ctx.user, job.id)
+
+    trxl = TransactionsLog.open('recommend_gender', job_ctx)
+
+    for rec in engine.recommend('gender', uuids, api_token):
+        results[rec.key] = {'gender': rec.options[0],
+                           'accuracy': rec.options[1]}
+
+    trxl.close()
+
+    logger.info(
+        f"Job {job.id} 'recommend gender' completed; "
+        f"{len(results)} recommendations generated"
+    )
+
+    return job_result
+
+
+@django_rq.job
 def affiliate(ctx, uuids=None):
     """Affiliate a set of individuals using recommendations.
 
--- a/sortinghat/core/jobs.py
+++ b/sortinghat/core/jobs.py
@@ -28,7 +28,7 @@ import django_rq.utils
 import pandas
 import rq
 
-from .api import enroll, merge
+from .api import enroll, merge, update_profile
 from .context import SortingHatContext
 from .errors import BaseError, NotFoundError, EqualIndividualError
 from .log import TransactionsLog
@@ -220,7 +220,7 @@ def recommend_matches(ctx, source_uuids, target_uuids, criteria, verbose=False):
 
 
 @django_rq.job
-def recommend_gender(ctx, uuids, api_token=None):
+def recommend_gender(ctx, uuids):
     """Generate a list of gender recommendations from a set of individuals.
 
     This job generates a list of recommendations with the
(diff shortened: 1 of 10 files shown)


tests/test_jobs.py
```diff
@@ -22,6 +22,8 @@
 
 import datetime
 import unittest.mock
+import json
+import httpretty
```
sduenas, line 26: `httpretty` is a third party library. We usually leave a blank line between standard libraries and third-party ones.

sortinghat/core/jobs.py
```diff
@@ -219,6 +219,49 @@ def recommend_matches(ctx, source_uuids, target_uuids, criteria, verbose=False):
     return job_result
 
 
+@django_rq.job
+def recommend_gender(ctx, uuids, api_token=None):
```
sduenas, line 223: I think the API token should be defined in the configuration of SortingHat instead of giving it in the call.

tests/rec/test_gender.py
```diff
@@ -0,0 +1,138 @@
...
+            individual does not have a valid name"""
+
+        setup_genderize_server()
+
+        uuids = [self.smith.uuid, self.no_name.uuid]
+        recs = list(recommend_gender(uuids))
+
+        self.assertEqual(len(recs), 0)
+
+    def test_not_found_individual(self):
```
sduenas, line 115: I think that this test should be also included in the tests for the mutations so we can check that everything goes ok.

sortinghat/core/jobs.py
```diff
@@ -421,6 +421,73 @@ def _group_recommendations(recs):
     return job_result
 
 
+@django_rq.job
+def autogender(ctx, uuids=None, api_token=None):
```
sduenas, line 425: I never liked this name. I think we should use `genderize` (which is linked to the service we use).

sduenas (changes requested):
The PR in overall looks great but some changes are needed. Also, it would be great if some of the tests that we have on the original code can be added too. For example, [test_retry](https://github.com/chaoss/grimoirelab-sortinghat/blob/master/tests/test_cmd_autogender.py#L395).

evamillan:
I updated the PR with the changes you requested.

tests/rec/test_gender.py
sduenas: We finally decided this wasn't needed because during the mutation tests we mock the response, so this function is not going to fail.

sortinghat/core/recommendations/gender.py
```diff
@@ -0,0 +1,138 @@
...
+
+import logging
+import re
+
+import requests
+import urllib3.util
+
+from ..db import find_individual_by_uuid
+from ..errors import NotFoundError, InvalidValueError
+from django.conf import settings
```
sduenas, line 33: I would move this inside `_genderize` function and get the `GENDERIZE_API_KEY` only there. That will mean that the key will be read again if it's updated, so we won't need to restart the system again if the file changes. Also, the value of the key in memory will be (or should be) destroyed once the function is done.

sortinghat/core/recommendations/gender.py
```diff
@@ -0,0 +1,138 @@
...
+
+logger = logging.getLogger(__name__)
+
+name_pattern = re.compile(r"(^\w+)\s\w+")
+
+api_key = settings.GENDERIZE_API_KEY
```
sduenas, line 39: Can we change the name of the setting to `SORTINGHAT_GENDERIZE_API_KEY`? I start thinking we should use `SORTINGHAT_` prefix with all the SortingHat settings to differentiate them from the rest of the settings.

sduenas (changes requested):
Good job. Just minor changes and please, can you squash some of the commits that have the same functionality?

sortinghat/core/recommendations/gender.py
sduenas: See my suggestion below. You can delete his line from above.

sortinghat/core/recommendations/gender.py
```diff
@@ -0,0 +1,137 @@
...
+    except Exception as e:
+        raise InvalidValueError(msg=str(e))
+    else:
+        return first_name
+
+
+def _genderize(name):
+    """Fetch gender from genderize.io"""
+
+    api_key = settings.SORTINGHAT_GENDERIZE_API_KEY
```
sduenas, line 101:
```suggestion
    from django.conf import settings
    
    api_key = settings.SORTINGHAT_GENDERIZE_API_KEY
```

sduenas (approved):
LGTM

PR #560, opened 2021-07-08
[ui] Get GraphQL API URL from env variable

### LINKED ISSUE #559: Graphql API URL as environment variable
The API URL used in `main.js` can change depending on the context, so it should be stored as an environment variable. The default value should be `/api`.

evamillan:
The default API URL `/api/` can be overwritten with the `VUE_APP_API_URL` environment variable. It can be set on an .env file or passed as a command line argument when the UI is launched, eg. `yarn serve --api_url=http://localhost:8000/api/`

Closes #559.


sduenas (approved):
LGTM

PR #562, opened 2021-07-12
[schema] Add GraphQL layer for teams

### LINKED ISSUE #558: Extend GraphQL layer to add queries and mutations for handling Teams
We need to create methods in the GraphQL layer to allow users to:

- [x]  add a team
- [x]  find a team - given an org, find all teams. 
- [x] delete a team

Rashmi-K-A:
This PR includes the changes for extending the GraphQL layer to manipulate Team objects. It adds a query for searching through Team objects and two mutations - `AddTeam` and `DeleteTeam` for creating and deleting Team objects respectively.

This PR closes #558.

### DIFF
--- a/sortinghat/core/schema.py
+++ b/sortinghat/core/schema.py
@@ -52,8 +52,10 @@ from .api import (add_identity,
                   merge,
                   unmerge_identities,
                   add_organization,
+                  add_team,
                   add_domain,
                   delete_organization,
+                  delete_team,
                   delete_domain,
                   enroll,
                   withdraw,
@@ -70,6 +72,7 @@ from .jobs import (affiliate,
                    recommend_gender,
                    genderize)
 from .models import (Organization,
+                     Team,
                      Domain,
                      Country,
                      Individual,
@@ -169,6 +172,11 @@ class TransactionType(DjangoObjectType):
 class OrganizationType(DjangoObjectType):
     class Meta:
         model = Organization
+
+
+class TeamType(DjangoObjectType):
+    class Meta:
+        model = Team
 
 
 class DomainType(DjangoObjectType):
@@ -291,6 +299,25 @@ class OrganizationFilterType(graphene.InputObjectType):
     term = graphene.String(
         required=False,
         description='Filter organizations whose name or domains include the term.'
+    )
+
+
+class TeamFilterType(graphene.InputObjectType):
+    name = graphene.String(
+        required=False,
+        description='Filter teams with an exact name match.'
+    )
+    organization = graphene.String(
+        required=False,
+        description='Filter teams belonging to this organization.'
+    )
+    parent = graphene.String(
+        required=False,
+        description='Filter teams which are subteams of parent.'
+    )
+    term = graphene.String(
+        required=False,
+        description='Filter teams whose name include the term.'
     )
 
 
@@ -434,6 +461,11 @@ class OrganizationPaginatedType(AbstractPaginatedType):
     page_info = graphene.Field(PaginationType, description='Information to aid in pagination.')
 
 
+class TeamPaginatedType(AbstractPaginatedType):
+    entities = graphene.List(TeamType, description='A list of teams.')
+    page_info = graphene.Field(PaginationType, description='Information to aid in pagination.')
+
+
 class IdentityPaginatedType(AbstractPaginatedType):
     entities = graphene.List(IndividualType, description='A list of identities.')
     page_info = graphene.Field(PaginationType, description='Information to aid in pagination.')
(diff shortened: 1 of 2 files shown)


sduenas (approved):
The PR looks good to me. Good job. Before we merging it, can you rebase it with muggle branch  (I merged your other PR before).

I also have another question and I don't know if we can fix it in this PR. Probably we can do it in a different one. The thing is, when we run a query to ask for Teams, the fields related to the django tree package are visible. I think those fields shouldn't be exposed with the api cause they are details of the implementation but maybe I'm wrong. Also, if I understand correctly, it's not possible to ask for the children of a team. That's something that we should do because it's part of the idea of GraphQL.

Rashmi-K-A:
> The PR looks good to me. Good job. Before we merging it, can you rebase it with muggle branch (I merged your other PR before).
> 
> I also have another question and I don't know if we can fix it in this PR. Probably we can do it in a different one. The thing is, when we run a query to ask for Teams, the fields related to the django tree package are visible. I think those fields shouldn't be exposed with the api cause they are details of the implementation but maybe I'm wrong. Also, if I understand correctly, it's not possible to ask for the children of a team. That's something that we should do because it's part of the idea of GraphQL.

@sduenas:

About the implementation details - I think am a bit confused here, it doesnt return any of the fields like path and depth and num children in the graphQL response. Did you mean something else?

About getting the children - I was thinking of this, it is just a different function call `getChildren` instead of `getDescendants`, maybe we can add a bool filter that can toggle based on whether users want just the children or all descendants. Or if you think we dont have a need for users to see all descendants, we can just use `getChildren` instead.

sduenas:
> > The PR looks good to me. Good job. Before we merging it, can you rebase it with muggle branch (I merged your other PR before).
> > I also have another question and I don't know if we can fix it in this PR. Probably we can do it in a different one. The thing is, when we run a query to ask for Teams, the fields related to the django tree package are visible. I think those fields shouldn't be exposed with the api cause they are details of the implementation but maybe I'm wrong. Also, if I understand correctly, it's not possible to ask for the children of a team. That's something that we should do because it's part of the idea of GraphQL.
> 
> @sduenas:
> 
> About the implementation details - I think am a bit confused here, it doesnt return any of the fields like path and depth and num children in the graphQL response. Did you mean something else?
> 

It does. For example, if you run this query:
```
query teams {
  teams {
    entities {
      name
      path
      numchild
    }
  }
}
```

I get the next results for the two teams that I created:
```
{
  "data": {
    "teams": {
      "entities": [
        {
          "name": "test",
          "path": "0001",
          "numchild": 0
        },
        {
          "name": "test2",
          "path": "0002",
          "numchild": 0
        }
      ]
    }
  }
}
```

This is because we're exposing these fields from the internal model. I'm not sure if it's possible to avoid it.

> About getting the children - I was thinking of this, it is just a different function call `getChildren` instead of `getDescendants`, maybe we can add a bool filter that can toggle based on whether users want just the children or all descendants. Or if you think we dont have a need for users to see all descendants, we can just use `getChildren` instead.

The descendants should be resolved within the query. I don't think we should add any filter to that. It should be something like:
```
query teams {
  teams {
    entities {
      name
      subteams {
         name
         subteams {
           name
           ...
      }
    }
  }
```

It's not mandatory to have it now but I think it's something we need because it's kind of the purpose of having graphql and a REST API, for example.

Rashmi-K-A:
> > @sduenas:
> > About the implementation details - I think am a bit confused here, it doesnt return any of the fields like path and depth and num children in the graphQL response. Did you mean something else?
> 
> It does. For example, if you run this query:
> 
> ```
> query teams {
>   teams {
>     entities {
>       name
>       path
>       numchild
>     }
>   }
> }
> ```
> 
> I get the next results for the two teams that I created:
> 
> ```
> {
>   "data": {
>     "teams": {
>       ...
>   }
> }
> ```
> 
> This is because we're exposing these fields from the internal model. I'm not sure if it's possible to avoid it.

@sduenas  Oh! I did not notice that. Can we not use `exclude=(path, depth,.....)` on the `TeamType` to exclude certain fields?

Rashmi-K-A:
> 
> > About getting the children - I was thinking of this, it is just a different function call `getChildren` instead of `getDescendants`, maybe we can add a bool filter that can toggle based on whether users want just the children or all descendants. Or if you think we dont have a need for users to see all descendants, we can just use `getChildren` instead.
> 
> The descendants should be resolved within the query. I don't think we should add any filter to that. It should be something like:
> 
> ```
> query teams {
>   teams {
>     entities {
>       name
>       subteams {
>          name
>          subteams {
>            name
>            ...
>       }
>     }
>   }
> ```
> 
> It's not mandatory to have it now but I think it's something we need because it's kind of the purpose of having graphql and a REST API, for example.


Sure, I think we can work out a way to do that. I ll make note of it for now and we can circle back to it when the UI is done.

sduenas (approved):
LGTM

sduenas:
I merged the PR but I'm going to open two issues to try to resolve the open questions of this PR.

PR #563, opened 2021-07-12
[api] Fix delete_team method

Rashmi-K-A:
Previously, the delete_team method could not handle deleting teams
that do not belong to any organization. This change removes the
requirement for delete_team to have organization as a parameter.

Signed-off-by: Rashmi-K-A <k.a.rashmi04@gmail.com>


sduenas (approved):
LGTM

PR #564, opened 2021-07-12
Add command 'sortinghatw' to run workers

sduenas:
The command added to this commit is a wrapper of the command 'rqworker'. Instead of running it with 'django-admin' or 'manage.py', we can use 'sortinghatw' to do it.


evamillan (approved):
LGTM

Issue #570, opened 2021-07-20
[gender] name pattern doesn't work if the name has one term


zhquan:
The [name_pattern](https://github.com/chaoss/grimoirelab-sortinghat/blob/muggle/sortinghat/core/recommendations/gender.py#L36) used on [_get_individual_name](https://github.com/chaoss/grimoirelab-sortinghat/blob/muggle/sortinghat/core/recommendations/gender.py#L85) method works if the name has two or more terms. But if the name has only one term or if the second term is separated by more than 1 blank space, it doesn't work.

Regular expression: `(^\w+)\s\w+`
Examples:

1. `Tom`: The regular expression does not match
2. `Tom  Jerry` (With 2 blank spaces): The regular expression does not match
3. `Tom Jerry`:
    - Match 1: `Tom Jerry`
    - Group 1: `Tom`
4. `Tom Jerry House`:
    - Match 1: `Tom Jerry`
    - Group 1: `Tom`

This means the `genderize` and `recommendGender` mutations do not work for cases 1 and 2.

evamillan:
Should it work for names that only have one term? It looks like it was a deliberate choice to not match them seeing [this commit](https://github.com/chaoss/grimoirelab-sortinghat/commit/2915cc456a978dcf6cde323afed7cc0a6a36394d) from the previous version:

```
[cmd:autogender] Check well-formed names only

Only those profiles that follow the pattern 'name surname/s' will
be suitable for assigning a gender.
```

sduenas:
I think we decided that to avoid problems with names that are nicknames, like `sduenas`, or `root`, for example. Than means that in big databases, the worker will need to go too many times to the service when it's not really needed.

Maybe we can do it better. For example, we can add a flag that depending on whether is set or not it will check if the name matches the pattern or not. Also, we should check that the name has at least 2 characters. We also can cache the results using [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) during the process to request names that there were already requested.

What do you think?

PR #574, opened 2021-08-02
[gender] Improve name patterns

evamillan:
This PR implements the suggestions on issue #570, differentiating between strict and loose name matching and caching the recommendation results. By default, the name validation will check for name following a `Name Lastname` pattern, with any number of spaces between them, and for the name to be at least two characters long. This can be disabled with the `no_strict_matching` flag, which will only check that the name has a minimum of 2 characters.

### DIFF
--- a/tests/rec/test_gender.py
+++ b/tests/rec/test_gender.py
@@ -95,23 +95,32 @@ class TestRecommendGender(TestCase):
         self.jane_doe = api.add_identity(self.ctx,
                                          name='Jane Doe',
                                          source='scm')
-        self.smith = api.add_identity(self.ctx,
-                                      name='Smith',
-                                      source='scm')
+        self.john = api.add_identity(self.ctx,
+                                     name='John',
+                                     source='scm')
         self.no_name = api.add_identity(self.ctx,
                                         email='email@example.com',
                                         source='scm')
+        self.double_space = api.add_identity(self.ctx,
+                                             name='John  Smith',
+                                             source='scm')
+        self.initial = api.add_identity(self.ctx,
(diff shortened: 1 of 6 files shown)


tests/rec/test_gender.py
```diff
@@ -95,23 +95,29 @@ def setUp(self):
...
     @httpretty.activate
-    def test_recommend_gender(self):
-        """Check if it returns a gender for valid names"""
+    def test_recommend_gender_strict(self):
+        """Check if it returns a gender for strict valid names"""
 
         setup_genderize_server()
 
-        uuids = [self.john_smith.uuid, self.jane_doe.uuid]
+        uuids = [self.john_smith.uuid, self.jane_doe.uuid, self.double_space.uuid]
```
sduenas, line 117: Can we add here the `self.initial` individual and then we can test nothing is returned for that individual? We can also added to line 167.

sduenas (changes requested):
LGTM. I'd add just two minor changes, though.

tests/rec/test_gender.py
evamillan: I'm not sure it's needed, `test_invalid_name_not_strict` and `test_invalid_name_strict` already check that nothing is returned for `self.initial`.

sduenas: I know. It's just to check that there's no a side effect when an invalid name is passed with valid names. In those tests we only pass one value.

evamillan:
I updated the PR with the changes.

sduenas (approved):
LGTM

PR #575, opened 2021-08-02
[docs] Add basic config files for readthedocs

VSevagen:
This commit adds the basic files needed to setup
readthedocs with sortinghat. The required dependencies
are added and the lock file is updated.

Signed-off-by: sevagenv <sevagenv@gmail.com>


.readthedocs.yml
```diff
@@ -0,0 +1,20 @@
+---
+version: 2
+
+build:
+  image: testing
+  apt_packages:
+    - python3-dev
+    - libmysqlclient-dev
```
sduenas, line 8: Why do we need this package in the configuration of `readthedocs`?

docs/conf.py
```diff
@@ -0,0 +1,111 @@
...
+    'default': {
+        'HOST': 'localhost',
+        'PORT': 6379,
+        'ASYNC': True,
+        'DB': 0
+        }
+    },
+    DEFAULT_GRAPHQL_PAGE_SIZE = 2
+)
+django.setup()
```
sduenas, line 40: Again, I don't know why we need this.

sduenas (changes requested):
Thanks for the patch. I have some questions about why MySQL packages and django configuration is needed.

docs/conf.py
VSevagen: The readthedocs build will leave some warning and some of the developer docs will be missing. Those configs are needed by the build to generate the docs properly.

.readthedocs.yml
VSevagen: There were some issues with `mysqlclient`. The readthedocs build would fail and the error message was that the version number does not exist, which is not true. I looked it up and it seems like the build environment did not include the mysql development libraries which is `libmysqlclient-dev`. Regarding `python3-dev` it might not be needed, but it was included in the instructions.

docs/conf.py
sduenas: Ok, well, let's keep it this way, then. I guess is because sphynx imports the packages and modules to get the docstrings, classes, etc.

README.md
```diff
@@ -245,4 +245,4 @@ $ yarn test:unit
 
 ## License
 
-Licensed under GNU General Public License (GPL), version 3 or later.
+Licensed under GNU General Public License (GPL), version 3 or later.
```
sduenas, line 248: Why does this line change?

.readthedocs.yml
sduenas: Ok. Same issue because it needs to import the modules to generate the docs.

docs/index.rst
```diff
@@ -0,0 +1,15 @@
+.. documentation documentation master file, created by
```
sduenas, line 1: This file needs to be in `rst` while we have the others in `Markdown`?

poetry.lock
```diff
@@ -1,3 +1,11 @@
+[[package]]
```
sduenas, line 1: I see many changes here that I think they aren't needed. Generate this file using `poetry lock --no-update`

sduenas (changes requested):
You resolved my questions but I opened some more.

docs/index.rst
VSevagen: Yes. We can't define a table of content in markdown because readthedocs and sphinx do not understand it. So for every file where I need a table of content, I will be defining them as `rst`.

README.md
VSevagen: I frankly have no idea why this is showing as changed.

VSevagen: i think it must be because the added tag doesn't count as an additional line, like total number of lines didn't change according to github but the added does increase the number of lines by one.

docs/README.md
```diff
@@ -0,0 +1,8 @@
+# How to generate this document locally
+
+- Clone this repository and install the required dependencies through poetry.
```
sduenas, line 3: I think you should add the command needed to to this because the dependencies for building the docs are defined as optional.

docs/README.md
```diff
@@ -0,0 +1,8 @@
...
+- Activate the virtual environment and change your directory to <code>docs</code>
+- Install sphinx and furo through the following commands<br>
+  <code>pip install sphinx</code><br>
```
sduenas, line 6: I think we should avoid HTML tags on markdown files whenever is possible. Code tag can be replaced by code markdown blocks. You don't need to use <br> if you use the blocks instead.

docs/README.md
sduenas: Also, not sure why we need pip to install those dependencies if they can be installed with poetry.

sduenas (changes requested):
I was able to generate the docs but I got a long warning:

```
WARNING: __all__ should be a list of strings, not [<class 'sortinghat.cli.client.client.SortingHatClient'>, <class 'sortinghat.cli.client.client.SortingHatClientError'>, schema 
```

The rest of the PR looks ok but you should rebase it to the latest commit of muggle. Thanks!

poetry.lock
VSevagen: I updated the lock file

docs/README.md
VSevagen: ok. It makes sense. I updated the docs with the appropriate instructions.

docs/README.md
VSevagen: I updated it. Thank you for the mention.

sduenas (approved):
LGTM. Thanks!

VSevagen:
> WARNING: __all__ should be a list of strings, not [<class 'sortinghat.cli.client.client.SortingHatClient'>, <class 'sortinghat.cli.client.client.SortingHatClientError'>, schema

This can be fixed. The items in `__all__`  should be converted to string but I think this would affect sortinghat leaving aside readthedocs.

PR #576, opened 2021-08-04
[docs] Add initial content for documentation

VSevagen:
This commit add the initial content, including the developer
docs. The section related to the use of SortingHat will
be added on the next pr.

Signed-off-by: sevagenv <sevagenv@gmail.com>


sduenas (changes requested):
For now it looks ok but I have a general comment. Can we set a maximum line length. Like 90 or something. Also without whitespaces at the end of the line? I'm afraid lines can be very long son it's better to have that. What do you think?

VSevagen:
> For now it looks ok but I have a general comment. Can we set a maximum line length. Like 90 or something. Also without whitespaces at the end of the line? I'm afraid lines can be very long son it's better to have that. What do you think?

I agree. Lines will definitely get longer. I'll fix this pr and the other docs as well.

VSevagen:
I did not change `index.rst` as its table of content needs the current syntax to work properly. I tried altering its content to fit the bill but the generated document was faulty.

sduenas (approved):
LGTM

sduenas:
Would you mind to rebase your branch to the latest changes before we merge it?

VSevagen:
Rebased it. Thank you for reviewing.

Issue #569, opened 2021-07-19
Hide implementation details in teams model


sduenas:
Some implementation details about how relationships between teams are stored in the database are exposed using the GraphQL API. End users don't need this information so it shouldn't be available.

For instance, fields such as `path` should't be accesible with the following query:

```
query teams {
  teams {
    entities {
      name
      path
    }
  }
}

{
  "data": {
    "teams": {
      "entities": [
        {
          "name": "test",
          "path": "0001"
        },
        {
          "name": "test2",
          "path": "0002"
        }
      ]
    }
  }
}
```

We need to determine which fields should be available and which not and block their access.

evamillan:
Those fields are already excluded from the [Team type](https://github.com/chaoss/grimoirelab-sortinghat/blob/91fe8f6d9bd458b61d13354f3a0657974cea21ea/sortinghat/core/schema.py#L209).

PR #582, opened 2021-08-10
[ui] add TeamModal component

Rashmi-K-A:
This commit adds a modal component to the Organizations
table. On clicking the View teams button in the table,
this modal opens up displaying all the top level teams
of the organization.

Signed-off-by: Rashmi-K-A <k.a.rashmi04@gmail.com>

### DIFF
--- a/ui/src/components/TeamModal.vue
+++ b/ui/src/components/TeamModal.vue
@@ -1,0 +1,247 @@
+<template>
+  <v-dialog v-model="isOpen" persistent max-width="1000px">
+    <v-card class="section">
+      <v-card-title class="header">
+        <span class="title"> {{ organization }} teams </span>
+      </v-card-title>
+      <form>
+        <v-card-text class="team-modal-content">
+          <v-treeview
+            :items="teams"
+            :load-children="getSubTeams"
+            :open.sync="open"
+            color="warning"
+            open-on-click
+            transition
+          >
+            <template v-slot:append="{ item }">
+              <v-row class="align-center">
+                <v-edit-dialog @save="add(item)">
+                  <v-icon>
+                    mdi-plus-circle-outline
+                  </v-icon>
+                  <template v-slot:input>
+                    <v-text-field
+                      v-model="teamName"
+                      label="Add a new team"
+                      maxlength="50"
+                      single-line
+                    />
+                  </template>
+                </v-edit-dialog>
+                <v-btn icon @click="confirmDelete(item)">
+                  <v-icon>
+                    mdi-delete
+                  </v-icon>
+                </v-btn>
+              </v-row>
+            </template>
+          </v-treeview>
+        </v-card-text>
+        <v-card-actions>
+          <v-row>
+            <v-col cols="3">
+              <v-edit-dialog @save="add()">
+                <v-btn text small left outlined color="primary">
+                  <v-icon small color="primary">mdi-plus-circle-outline</v-icon>
+                  Add team
+                </v-btn>
+                <template v-slot:input>
+                  <v-text-field
+                    v-model="teamName"
+                    label="Add a new team"
+                    maxlength="50"
+                    single-line
+                  />
+                </template>
+              </v-edit-dialog>
+            </v-col>
+          </v-row>
+          <v-spacer></v-spacer>
+          <v-btn color="primary darken-1" text @click.prevent="closeModal">
+            Close
+          </v-btn>
+        </v-card-actions>
+      </form>
+    </v-card>
+    <v-dialog v-model="dialog.open" max-width="500px">
+      <v-card class="pa-3">
+        <v-card-title class="headline">{{ dialog.title }}</v-card-title>
+        <v-card-text>
+          <p class="pt-2 pb-2 text-body-2">{{ dialog.text }}</p>
+        </v-card-text>
+        <v-card-actions v-if="dialog.action">
+          <v-spacer></v-spacer>
+          <v-btn text @click="closeDialog">
+            Cancel
+          </v-btn>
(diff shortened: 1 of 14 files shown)


ui/src/components/TeamModal.vue
```diff
@@ -0,0 +1,271 @@
...
+        open: false,
+        title: '',
+        text: '',
+        action: '',
+      },
+    };
+  },
+  computed: {
+    items() {
+      return this.teams;
```
evamillan, line 163: I don't think you need this, you can use `this.teams` directly on the treeview.

ui/src/components/TeamModal.vue
```diff
@@ -0,0 +1,271 @@
...
+        open: false,
+        title: '',
+        text: '',
+        action: '',
+      });
+    },
+    async getTeams(filters = this.filters) {
+      const response = await this.fetchTeams(this.page, this.itemsPerPage, filters);
+      if (response) {
+        this.pageCount = response.data.teams.pageInfo.numPages;
```
evamillan, line 190: If there is no pagination we probably don't need all this data.

ui/src/components/TeamModal.vue
```diff
@@ -0,0 +1,271 @@
...
+          title: 'Error',
+          text: this.$getErrorMessage(error),
+          action: null,
+        });
+        this.$logger.error(`Error adding team ${this.teamName}: ${error}`);
+      }
+    },
+  },
+  async created() {
+    this.filters = { organization: this.organization };
```
evamillan, line 266: This can be moved to the `filters` initialization in `data()`.

ui/src/components/TeamModal.vue
```diff
@@ -0,0 +1,271 @@
...
+    const teamObjects = await this.getTeams();
+    teamObjects.forEach((team) => this.teams.push({ name: team.name, children: [] }));
```
evamillan, line 268: This looks a bit redundant, maybe adding one level of `subteams` to the query would make this initialization of `children` unnecessary and improve the UX. Right now, all teams have an arrow by their name, implying they have subteams, and when you click one that doesn't have children it looks like a bug. What do you think?

evamillan (commented):
Thanks for the PR, Rashmi, I've got a few comments.

- I get some warnings when I run `yarn serve`, running `yarn lint --fix` should fix most of them.
- Can you align the edit and delete buttons on the modal? Adding the `align-center` class to the row they're in will probably do the trick. Also, the "add a new team" button looks a bit like plain text, you can maybe use the same one as the "add domain" or "add organization" from the other modals.
- I thought we would display the number of teams and the top level ones on the organizations table, it could be useful to have that bit of information there. In any case, the expanded view can look tidier if the "Domains" and "Teams" text has the same size and is aligned.
- When I add a subteam on the modal, the team is added to the database but the interface is not updated, you need to reload the page to see it.

I also left some comments on the code.

ui/src/components/TeamModal.vue
Rashmi-K-A: It would help for just that one level, sure. But in the second level down there would still have to be the arrows because we dont know which of the children in the first level are leaf teams and which of them have subteams. The easiest way to make all of this simpler would be to just write a query for around five levels of subteams and construct the org tree on component creation. It would simplify a lot of the code in the rest of the functions as well. What do you think?

evamillan: You're right, it would only work for the first level, after that we'd have the same problem. What about using the `numchild` field for this? If it's other than 0, you can assume there are children and add the empty array then.
I agree that writing a query with multiple levels would simplify the code a lot, but we would have to set an artificial limit of levels somewhere, so I'm not sure about it.

Rashmi-K-A: I believe we are working on a ticket to hide those fields as they fall under implementation details. Check -  https://github.com/chaoss/grimoirelab-sortinghat/issues/569, but having numChildren might be useful. We could check with Santiago about exposing just that field but we dont actually have it as part of the Teams model.

evamillan: I think new subteams are not being updated properly because of the condition on the `add` function on line 244 `if (parent.children.length > 0)`. It only adds the child if there are other children.

Rashmi-K-A: otherwise if i dont have that condition , vue would  see that the children array is not empty so it would never call load children. none of the other subteams would show up

Rashmi-K-A: an alternate is to call get subteams as part of the call for add team for the parent passed

evamillan: > an alternate is to call get subteams as part of the call for add team for the parent passed

This could work, can you try it?

Rashmi-K-A: yep, sure can! And about using numchildren as a way of checking whether a team has any chlidren, where did we land on that? If you think its a good idea, I ll loop in Santiago to confirm from his side if its ok

evamillan: Yes, I think it's the best way to check for children if we don't want to limit the depth of subteams.

Rashmi-K-A: Sure I ll put in the changes

evamillan:
The changes look good to me, but there are still some lint warnings, can you take a look at them @Rashmi-K-A?

Rashmi-K-A:
> The changes look good to me, but there are still some lint warnings, can you take a look at them @Rashmi-K-A?

Done @evamillan

evamillan (approved):
LGTM

PR #583, opened 2021-08-26
[server] Create database during setup

sduenas:
With this patch, the command 'sortinghat-admin setup' creates an empty database before migrations and fixtures
are applied.

The new database will have the name of the variable 'NAME' under the default database settings.

This database will be only created if it doesn't exist.


zhquan (approved):
LGTM

PR #586, opened 2021-09-23
Handle Vue routes from Django urls

### LINKED ISSUE #585: Vue routes are not found on page refresh
If a user navigates to a page, i.e. `/login`, or refreshes the page, it shows a "Not Found" error. This doesn't happen if the route is loaded by clicking an internal link.

evamillan:
This PR replaces the empty `path` for the UI with a regular expression to capture Vue's routes.
Fixes #585 .

### DIFF
--- a/sortinghat/app/urls.py
+++ b/sortinghat/app/urls.py
@@ -2,7 +2,7 @@
 
 
 from django.conf import settings
-from django.urls import path
+from django.urls import path, re_path
 from django.views.generic import TemplateView
 
 from sortinghat.core.views import SortingHatGraphQLView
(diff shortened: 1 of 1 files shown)


sduenas (changes requested):
This doesn't work with the current configuration of SortingHat. The new pattern is always returning `index.html` even for static files. You can try with the following commands to run the server and accesing `http://localhost:8000`.
```
$ export SORTINGHAT_SECRET_KEY="test"

$ sortinghatd --config sortinghat.config.settings --dev 
```

evamillan:
I've changed it to exclude `/static`, it should work with the `sortinghatd` command now.

sduenas (approved):
LGTM

PR #590, opened 2021-11-19
[ui] Allow deploy SortingHat in a sub-path

zhquan:
This change allows us to deploy SortingHat UI in a sub-path
(https://localhost:8000/sortinghat) instead of root
(https://localhost:8000)


sduenas (approved):
LGTM

PR #595, opened 2022-01-03
[ui] Show unaffiliated teams

evamillan:
This PR adds a table for groups (teams that aren't affiliated to any organization) on the main view. It uses the `OrganizationTable` component since the functionality and look is almost identical, but adds the `isGroup` prop to manage the differences like hiding the domains or passing different parameters to queries and mutations.


ui/src/components/TeamModal.vue
```diff
@@ -189,7 +197,8 @@ export default {
     async delete(item) {
       this.closeDialog();
       try {
-        const response = await this.deleteTeam(item.name, this.organization);
+        const organization = this.isUnaffiliated ? null : this.parent;
```
sduenas, line 200: Shouldn't it be better to have `this.isAffiliated`? The logic might be clear but I'm not sure if we use `isUnaffiliated` somewhere else?

sduenas (commented):
LGTM, however check my comment about `unaffiliated`var.

ui/src/components/TeamModal.vue
evamillan: I've changed it to `isGroup`, which is the variable used in the other components.

sduenas (approved):
LGTM

PR #598, opened 2022-02-02
[model] Unify organizations and teams into Group

evamillan:
This PR adds a `Group` model that represents both organizations and teams, sharing the same database table. The `organizations` and `teams` tables are removed, but their models are kept as proxies of `Group`. These models set the type by default when objects are created and have custom query sets to filter them by type. 
In the GraphQL schema, `OrganizationType` and `TeamType` use the `Group` model, but some fields are hidden depending on their type.


tests/test_db.py
```diff
@@ -445,13 +438,38 @@ def test_add_organization(self):
         name = 'Example'
 
         org = db.add_organization(self.trxl, name)
-        self.assertIsInstance(org, Organization)
+        self.assertIsInstance(org, Group)
```
sduenas, line 441: Why did you change this test? I think it should keep testing that the object is an organization.

tests/test_db.py
```diff
@@ -445,13 +438,38 @@ def test_add_organization(self):
...
         org = Organization.objects.get(name=name)
         self.assertIsInstance(org, Organization)
         self.assertEqual(org.name, name)
 
+    def test_add_multiple_organizations(self):
+        """Check if more than one organization is added"""
+
+        name = 'Organization 1'
+        org = db.add_organization(self.trxl, name)
+        self.assertIsInstance(org, Group)
```
sduenas, line 454: Change these tests to check whether it is an `Organization` object or not.

tests/test_db.py
```diff
@@ -814,12 +828,12 @@ def setUp(self):
...
 
         self.trxl = TransactionsLog.open('delete_team', self.ctx)
-        self.org = Organization.objects.create(name='Example')
+        self.org = Organization.add_root(name='Example')
 
     def test_delete_team(self):
         """Check whether it deletes a team"""
 
-        team = Team.add_root(name='subTeam1', organization=self.org)
+        team = Group.add_root(name='subTeam1', organization=self.org, type='team')
```
sduenas, line 836: Shouldn't this be tested with `Team.add_root`?

tests/test_db.py
```diff
@@ -2039,30 +2053,30 @@ def test_add_multiple_enrollments(self):
         self.assertEqual(enrollment.end, datetime.datetime(2005, 1, 1, tzinfo=UTC))
         self.assertIsInstance(enrollment.individual, Individual)
         self.assertEqual(enrollment.individual.mk, mk)
-        self.assertIsInstance(enrollment.organization, Organization)
+        self.assertIsInstance(enrollment.organization, Group)
```
sduenas, line 2056: Check against `Organization`?

sortinghat/core/db.py
```diff
@@ -280,10 +281,13 @@ def add_organization(trxl, name):
 
     validate_field('name', name)
 
+    if Organization.objects.all_organizations().filter(name=name).exists():
```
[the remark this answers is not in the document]
sduenas: Also, we need a test for this.

sduenas (changes requested):
Can you check my comments, please? They are minor comments.

tests/test_db.py
evamillan: It returns a `Group` instead of an `Organization` because the foreign key is mapped to `Group`. Should I change it back to `Organization`?

sduenas: That's a good point. The enrollment is going to be linked to a group because we will have teams and organizations, so that might mean organization should be renamed to `group` in the future, isn't it?

evamillan: Yes, it will be a group so individuals can be enrolled in teams too.

sduenas: Ok, so, let's keep it this way for now but we will have to change the attribute `organization` to `group`, I think.

evamillan:
@sduenas I rebased the PR and updated it with the requested changes.

sortinghat/core/models.py
```diff
@@ -134,34 +134,73 @@ def __str__(self):
...
+        @classmethod
+        def choices(cls):
+            return ((item.name, item.value) for item in cls)
+
+        def __str__(self):
+            return self.value
+
     name = CharField(max_length=MAX_SIZE_CHAR_INDEX)
+    type = CharField(max_length=12, choices=GroupType.choices())
+    organization = ForeignKey('self', related_name='teams', on_delete=CASCADE,
```
sduenas, line 151: I'm not very sure about this field. It's for teams but not for organizations. The main problem is the name of the field in the table: `organization_id` which is not clear what its purpose is or what it stores.

sduenas (changes requested):
I've been testing the change and it works without problems. I have doubts about the field `organization_id` on `groups` table. Not sure if we should rename it or make more clear what it does.
In addition, I realized the script [utils/create_sh_0_7_fixture.py](`https://github.com/chaoss/grimoirelab-sortinghat/blob/muggle/utils/create_sh_0_7_fixture.py`) needs modifications to adapt it to the new schema. Would you mind to update it?

sduenas (approved):
LGTM

PR #602, opened 2022-02-23
Add poetry support to SortingHat

jjmerchante:
This PR:
- Adds the files required to use Poetry for building and installing the package.
- Moves the directories `bin/` and `misc/` inside SortingHat to be used in Poetry as scripts. 
- Changes `setup.py` to use the previous scripts as entry_points based on Perceval.


vchrombie:
The build is failing because of https://github.com/chaoss/grimoirelab-sortinghat/issues/603 and we have a PR to fix it https://github.com/chaoss/grimoirelab-sortinghat/pull/604

pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+keywords = [
+    "development",
+    "grimoirelab"
+]
+
+packages = [
+    { include = "sortinghat" },
+    { include = "tests", format = "sdist" },
+]
+
```
vchrombie, line 24: I think we missed this
```
include = [
    { path = "AUTHORS", format = "sdist" },
    { path = "NEWS", format = "sdist" },
    { path = "README.md", format = "sdist" },
]
```
https://github.com/chaoss/grimoirelab-toolkit/blob/a1fed329d25731177105fa094fa22b1029fdac1e/pyproject.toml#L25
https://github.com/chaoss/grimoirelab-toolkit/issues/44

Looks like we missed it in other tools as well

sduenas:
> The build is failing because of #603 and we have a PR to fix it #604

I've just merged the PR. @vchrombie do you think this PR looks good?

vchrombie (commented):
I had one minor comment, rest all looks good. I think we need to fix the build first before merging this.

cc @sduenas

vchrombie:
I see the PR https://github.com/chaoss/grimoirelab-sortinghat/pull/604 is merged, it should fix the build failure :+1:

pyproject.toml
vchrombie: @sduenas WDYT?

sduenas: Sure, we should add those. Where else did we miss to include those files?

vchrombie: I don't see the configurations in the perceval repo (I just opened a PR https://github.com/chaoss/grimoirelab-perceval/pull/776 for it)
I will check if the other tools also miss this, and will open the PRs accordingly.

vchrombie (approved):
LGTM! Thanks for working on this jj!

Issue #603, opened 2022-02-23
Jinja 2 version conflicts with the latest version of MarkupSafe


jjmerchante:
Installing Sortinghat using Jinja 2.11.3 conflicts with the latest version of MarkupSafe and throws the following error:

```ImportError: cannot import name 'soft_unicode' from 'markupsafe'```.

This is a known issue in Jinja and they propose to update to v3 (https://github.com/pallets/jinja/issues/1585).

We can pin Jinja to 3.0.3 because 3.1.0 will drop support for Python 3.6.

jjmerchante:
Fixed by #604

PR #604, opened 2022-02-23
[deps] Update Jinja version

jjmerchante:
Markup has release a new version and Jinja 2.11.3 throws an error.

Pin the version to <3.1.0 since from 3.1 the support for python 3.6 will be discontinued.

Related to https://github.com/chaoss/grimoirelab-sortinghat/issues/603


sduenas (approved):
LGTM

PR #609, opened 2022-03-08
[setup.py] Add missing commas in entry_points

jjmerchante:
This PR adds missing commas in entry_points command list.


zhquan (approved):
LGTM

PR #613, opened 2022-04-06
[muggle] Create release file and update packages

jjmerchante:
- Create the release file for muggle branch (GitHub Action)
- Remove Python 3.6 version
- Update Jinja2 version to v3


.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          artifact-name: sortinghat-dist
+          artifact-path: dist
+
+  tests:
+    needs: [build]
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql:5.7
```
sduenas, line 25: Can we use mariadb instead?

sduenas (changes requested):

.github/workflows/release.yml
jjmerchante: I changed the version to `mariadb:10.5`, do you prefer any other version of mariadb?

sduenas: That one is fine. Thanks!

sduenas (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -0,0 +1,88 @@
+name: release-muggle
+
+on:
+  push:
+    tags:
+      - '*.*.*'
+      - '*.*.*-*'
```
sduenas, line 7: For now, I think we should filter this by branch. Check this: https://github.com/chaoss/grimoirelab-sortinghat/blob/muggle/.github/workflows/tests.yml

sduenas (changes requested):
Sorry, we need a final change to be sure that this only works on muggle.

.github/workflows/release.yml
sduenas: Apparently, if we add the branch, it will create a release every time there's a commit to a branch. We have been checking and this workflow will only work for commits in muggle branch.

sduenas (approved):
LGTM

PR #615, opened 2022-04-20
[gitdm] Skip invalid format lines

zhquan:
This change skips invalid formatting lines instead of raising an
error. This way it will keep executing the file.

Example of invalid lines:
- bitergia.com	B#itergia
- jsmith.example.com  Example#
- jsmith.example.com  Example Company < 2010-01-01#

Test updated accordingly.

### DIFF
--- a/releases/unreleased/[gitdm]-skip-invalid-format-lines.yml
+++ b/releases/unreleased/[gitdm]-skip-invalid-format-lines.yml
@@ -1,0 +1,8 @@
+---
+title: '[gitdm] Skip invalid format lines'
+category: fixed
+author: Quan Zhou <quan@bitergia.com>
+issue: null
+notes: >
+    Gitdm parser won't fail reading files with an invalid format. Instead,
+    it will ignore invalid content.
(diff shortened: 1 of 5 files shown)


sduenas (changes requested):
I think the commit what really does is to skip those lines with an invalid format. You should reword the commit message an the changelog entry (also the filename). Besides that, the PR looks ok to me.

releases/unreleased/[gitdm]-skip-invalid-format-lines.yml
```diff
@@ -0,0 +1,8 @@
+---
+title: '[gitdm] Skip invalid format lines'
+category: fixed
+author: Quan Zhou <quan@bitergia.com>
+issue: null
+notes: >
+    Skip lines with invalid formatting instead of raising an error.
+    This way it will continue executing the file.
```
Comment (sduenas) [inline releases/unreleased/[gitdm]-skip-invalid-format-lines.yml:8]:
```suggestion
    Gitdm parser won't fail reading files with an invalid format. Instead, it will ignore
    invalid content
```

sduenas (changes requested):
LGTM but let me know what you think about my change on the changelog entry.

sduenas (approved):
LGTM

PR #629, opened 2022-07-20
[poetry] Allow Grimoirelab prereleases in dependencies

jjmerchante:
This PR updates the dependencies for this package and includes the `allow-prerelease` option for all grimoirelab dependencies.


jjmerchante:
I had to update all the packages (`poetry update`)  because the was an error for Python 3.7 and Pandas

PR #630, opened 2022-07-28
[ui] Create view to show an individual's info

### LINKED ISSUE #628: Separate view for individual's info
There should be a separate view that allows users to access a specific individual's data (profile, identities and enrollments).

evamillan:
This PR adds a view at `/individual/uuid` to show an individual's profile, identities and enrollments. A link to this view can be found in each individual's entry on the table, both in the buttons next to the name and in the dropdown menu.

imagen

imagen

The list of identities in the expanded individual and the modal to enroll in an organization in the table and workspace are moved to new components so that they can be also used in this view. It also adds a button on the enrollment list to add one without using drag and drop.

Closes #628.

### DIFF
--- a/ui/src/apollo/mutations.js
+++ b/ui/src/apollo/mutations.js
@@ -164,6 +164,12 @@ const ENROLL = gql`
           name
           email
           id
+          isBot
+          gender
+          country {
+            code
+            name
+          }
         }
         enrollments {
           start
--- a/ui/src/apollo/mutations.js
+++ b/ui/src/apollo/mutations.js
@@ -176,6 +176,10 @@ const ENROLL = gql`
           end
           group {
             name
+            type
+            parentOrg {
+              name
+            }
           }
         }
       }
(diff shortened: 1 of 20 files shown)


sduenas (changes requested):
The PR looks ok. However, I found a problem with the navigation. When I move to an individual view the history doesn't seem to be updated on the browser, so I can't go back to the main page.
Can you have it a look?

sduenas (changes requested):
We have decided that the icon for editing the name and email on the individual list should be removed because the full editing of the profile is going to be on the individual view.

Also, the icon "View full profile" should be also removed because the behavior will be to open the individual view when clicking on the individual's name.

sduenas (changes requested):
The view should also show the main UUID (i.e. mk field) of the individual.

evamillan:
I added the UUID to the view, removed the editable name and email fields from the table and replaced the icon with a link in the individual's name. We couldn't replicate the problem with the navigation, so nothing was changed in this regard.

evamillan:
I updated the PR to fix a bug that made the organizations data disappear when the individual was enrolled in a team.

jjmerchante:
LGTM Thank you!

zhquan (approved):
LGTM

PR #647, opened 2022-08-25
[ui] Resize search bar

### LINKED ISSUE #640: Resize search bar on individuals list
The length of the search bar should be longer to include more filters. Now it's too short so when two or three filters are added the rest of the parameters are not shown.

imagen

evamillan:
This PR resizes the search bar on the individual's table to take up all available space. It also reorders the table's actions items on smaller device sizes so the search bar still has enough space.

1920x1080:
imagen

1366x768:
imagen

Closes  #640.

### DIFF
--- a/ui/src/components/IndividualsTable.vue
+++ b/ui/src/components/IndividualsTable.vue
@@ -35,7 +35,7 @@
       >
       </v-checkbox>
       <search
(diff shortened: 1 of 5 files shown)


jjmerchante:
LGTM

PR #648, opened 2022-08-29
[ui] Click to copy UUID

### LINKED ISSUE #638: Copy Identity UUID
To copy the UUID, that it shown in the chip component of each identity, the user needs to select the text on it, and then, copy it either with the keyboard or opening the browser window. The problem is that the hash needs to be selected and also it isn't clear if it's going to copy only the part that is shown in the chip or the full UUID.

evamillan:
Adds an icon and explanatory text to the UUID chip in each identity to hint that the text can be copied.
The full UUID is copied on click and not only the visible part.

copy

Closes #638.

### DIFF
--- a/ui/src/components/Identity.vue
+++ b/ui/src/components/Identity.vue
@@ -1,13 +1,21 @@
 <template>
   <v-row no-gutters>
     <v-col class="uuid">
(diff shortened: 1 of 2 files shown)


jjmerchante:
LGTM

PR #649, opened 2022-08-30
[ui] Remove profile data from the Individuals table

### LINKED ISSUE #633: Remove profile data from the individuals list
Due to there's going to be a view that shows the full profile of an individual, we can remove the profile section from the individuals list.

evamillan:
Removes the "Profile" section for the expanded individuals on the table. Fixes #633.


jjmerchante:
LGTM

PR #651, opened 2022-09-05
[ui] Consistent margins for Individuals table entries

### LINKED ISSUE #635: Different top/bottom margins for entries in the Individual list
The margins for the Individuals entries are different when the entry has an organization than when it doesn't.
Margins should always has the size of individuals with organizations.

imagen

evamillan:
This PR sets the margin on the entry avatar so the size does not depend on the individual having an organization. It also removes the space reserved for the organization if there isn't one to keep the items aligned.
Fixes #635.

### DIFF
--- a/ui/src/components/IndividualEntry.vue
+++ b/ui/src/components/IndividualEntry.vue
@@ -18,7 +18,7 @@
   >
     <td width="25%">
       <v-list-item>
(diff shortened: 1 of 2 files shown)


jjmerchante:
LGTM Thank you!

jjmerchante:
LGTM Thank you!

PR #662, opened 2022-10-21
[core] Add permissions to execute jobs

### LINKED ISSUE #660: Add permissions to execute jobs
Jobs execution should be limited to admin users. We should define permissions for users so we can restrict who executes what. Take into account we're using `django auth` app, so we should reuse anything from there.

evamillan:
This PR adds an `execute_job` custom user permission and requests it for all GraphQL mutations that run jobs. Django's permissions are tied to models' content types, so a `CustomPermissions` dummy model with no database is created to store it.

The PR also includes a command to create an `Admins` group with all current permissions. The command can be run with `django-admin` or `manage.py`:
```
./manage.py create_groups --settings=sortinghat.config.settings
```

Fixes #660.


sortinghat/core/decorators.py
```diff
@@ -30,3 +30,9 @@
 check_auth = user_passes_test(
     lambda u: u.is_authenticated or not settings.SORTINGHAT_AUTHENTICATION_REQUIRED
 )
+
+
+def check_permissions(perms):
+    def test(user):
+        return user.has_perms(perms) or not settings.SORTINGHAT_AUTHENTICATION_REQUIRED
+    return user_passes_test(test)
```
sduenas, line 38: You can define this in the same way `check_auth` was defined.

sortinghat/core/management/commands/create_groups.py
```diff
@@ -0,0 +1,90 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2014-2021 Bitergia
```
sduenas, line 3:
```suggestion
# Copyright (C) 2014-2022 Bitergia
```

sortinghat/core/management/commands/create_groups.py
```diff
@@ -0,0 +1,90 @@
...
+
+import logging
+
+from django.core.management import BaseCommand
+from django.contrib.auth.models import Group, Permission
+from django.contrib.contenttypes.models import ContentType
+
+logger = logging.getLogger(__name__)
+
+GROUPS = {
```
sduenas, line 30:
```suggestion
SORTINGHAT_PERMISSION_GROUPS = {
```

sduenas (changes requested):
In overall, it looks good to me but I left some suggestions open.

sduenas (approved):
LGTM

PR #665, opened 2022-10-26
[muggle] Update workflow and remove unused files replaced by pyproject

jjmerchante:
This PR:
- Updates the workflows to include the SHA version of the action used
- Delete unused files replaced by `pyproject.toml` like requirements.txt and setup.py
- Update dependencies to match other GrimoireLab components

Some notes about dependencies:
- `setuptools` is now explicit because `uri` package depends on it but it is not in its dependencies. Latest version of Poetry set `setuptools` as optional.
- httppretty > 1.1 contains a bug that causes a test to fail. (Duplicate requests in latest_requests() if there are chunks)


sduenas (approved):
LGTM

PR #668, opened 2022-11-07
[dependency] Fix JWT dependency to <2

jjmerchante:
This PR fixes the JWT dependency to <2 because newer versions are not compatible with `django-graphql-jwt` and the latest version of that package includes a bug.


sduenas (approved):
LGTM

PR #669, opened 2022-11-08
Remove 'uri' package dependency

sduenas:
The uri package is not maintained and it doesn't have support for Python 3.10 or later.


jjmerchante:
LGTM

PR #670, opened 2022-11-08
Fix UTF-8 conversion on client

sduenas:
UTF-8 characters sent by the client weren't converted propertly due to an option in the JSON parser of the standard library.

By default, `json.dumps` ensures that everything is converted to ascii characters. This was problematic for emotions or other UTF-8 characters of 4 bytes. Setting the property `ensure_ascii` to `False` fixes this issue.

Tests have been updated accordingly.


jjmerchante:
LGTM

PR #673, opened 2022-11-22
Create queries and mutations for recommendations

jjmerchante:
Include new models for AffiliationRecommendation, MergeRecommendation,
and GenderRecommendation to store the results for the recommendation jobs.

Add mutations to obtain the recommendations and to apply them using GrahpQL.


evamillan:
The PR works well, thanks @jjmerchante . Just one small thing, the recommendations still show up on the individuals query when they've been applied, I don't know if they should be filtered out.

sduenas (approved):
LGTM

PR #674, opened 2022-11-29
Use Cypress for end to end testing

evamillan:
This PR replaces Puppeteer and Jest with Cypress for the end to end tests.

Run tests in headless mode with `yarn test:e2e` or use `yarn vue-cli-service test:e2e` to launch Cypress' GUI.
Environment variables `CYPRESS_USERNAME` and `CYPRESS_PASSWORD` are required to log in.
The default URL is http://localhost:8000, but it can be changed with the `--url` option, and the API URL with the `CYPRESS_API_URL` env variable.

To make the tests more independent from each other, the user is logged in using GraphQL instead of relying on the login test for all the following ones and new individuals are added for each test instead of using the same ones in all of them.


evamillan:
I made some changes to the PR:
- Installed Cypress directly instead of through a vue-cli plugin to fix an issue with the default URL. To change the URL, you can now use the `CYPRESS_BASE_URL` environment variable. To open Cypress' interface use `yarn cypress open`.
- All environment variables are in uppercase.
- Wait for a small amount of time after sorting the table to ensure it's interacting with the correct individual.

evamillan:
I updated the PR to add all the individuals before the tests and remove them after they are done, check that the token is returned and the username is shown on the header on the login test and fixed the API_URL variable.

The base URL can't be set with the `--env` option, only with `--config`. Eg. `yarn test:e2e --env USERNAME=xxx,PASSWORD=xxx --config baseUrl=xxx`.

PR #679, opened 2023-01-03
[ui] Fix the endpoint of the SortingHat API in production

sduenas:
For production deployments, the API of SortingHat will be under '/identities' path. Therefore, the path of the API will be '/identities/api'.

This is a temporary solution. We will try to make the deployment of SortingHat more flexible, so instances can use different paths and configurations where the SortingHat service will running.

Signed-off-by: Santiago Dueñas <sduenas@bitergia.com>


zhquan (approved):
LGTM

PR #680, opened 2023-01-04
Move generate_uuid method to sortinghat.utils

jjmerchante:
This PR moves the generate_uuid method from sortinghat.core.api to sortinghat.utils module


sduenas (approved):
LGTM

PR #682, opened 2023-01-05
Include static files in Python package

jjmerchante:
This commit includes the JS files and templates generated with yarn build in the Python package generated with Poetry.

It also update the httppretty package to ^1.1.4 because with "<1.1", peodd fails


.github/workflows/release.yml
```diff
@@ -11,6 +11,16 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # 3.1.0
+      - name: Setup Node ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
```
sduenas, line 16: This action should have a hash instead of a number version.

sduenas (changes requested):
The PR looks ok in overall. Please change the version number of the action before we merge it.

sduenas (approved):
LGTM

PR #719, opened 2023-01-10
[ui] Install Storybook

evamillan:
Adds Storybook as a dev dependency.


evamillan:
I bumped the Vue version and required Storybook to use the right template compiler and it seems to be working now.

sduenas (approved):
LGTM

PR #724, opened 2023-01-19
[poetry] Fix rq version and update fakeredis

jjmerchante:
This PR fixes the version of `rq` to `<1.12` because at `1.12.0` it raises the following error in tests:
`redis.exceptions.ResponseError: unknown command 'xrevrange'`

Version 1.12.0 of `rq` might be incompatible with `fakeredis`. In this PR, `fakeredis` is updated to `2.2.0`, the latest compatible version with Python 3.7.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -71,18 +71,101 @@ files = [
 
 [[package]]
 name = "charset-normalizer"
-version = "2.1.1"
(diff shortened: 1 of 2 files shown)


sduenas (approved):
LGTM

PR #725, opened 2023-01-20
Update setuptools version to >65.5.0

jjmerchante:
This commit updates the version of `setuptools` without restrictions to be compatible with other packages that has the same constraints.


sduenas (approved):
LGTM

PR #728, opened 2023-01-25
[ui] Fix routing issues

evamillan:
Fixes the issues with the individuals profile page not loading and the URL appending "/login" every time the page is refreshed


sduenas (approved):
LGTM

Issue #726, opened 2023-01-23
Migration to new database on docker


sduenas:
When the docker image runs, it needs to check whether the database is old (<=0.8) or not. If it has the old schema, the migration should be activated and import the new schema and the old data into it. This also means the script that produces the fixture for the new database needs to be updated.

PR #731, opened 2023-01-27
Migrate SortingHat 0.7 schema on Docker startup

jjmerchante:
This commits includes a new command in sortinghat-admin command to migrate the old database schema from SortingHat 0.7 to the new 0.8 version.

If the database schema is old, it creates a migration of the data, moves the database to a new database with `_backup`, initializes the database, and restores the data using the new schema.

Closes https://github.com/chaoss/grimoirelab-sortinghat/issues/726

### DIFF
--- a/sortinghat/server/sortinghat_admin.py
+++ b/sortinghat/server/sortinghat_admin.py
@@ -76,6 +76,10 @@ def setup(no_interactive, only_ui):
     It the flag 'no-interactive' was given, these environment
     variables are mandatory.
     """
+    _setup(no_interactive=no_interactive, only_ui=only_ui)
+
+
+def _setup(no_interactive, only_ui):
     env = os.environ
     env_vars = False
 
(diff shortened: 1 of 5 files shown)


sortinghat/server/sortinghat_admin.py
```diff
@@ -140,14 +144,86 @@ def upgrade(no_database):
...
+                cursor.execute(
+                    f"RENAME TABLE {from_db}.{table} TO {to_db}.{table};"
+                )
+        except MySQLdb.DatabaseError as exc:
+            msg = f"Error in backup database '{from_db}': {exc}."
+            raise click.ClickException(msg)
+
+    db_params = settings.DATABASES['default']
+
+    click.secho("Migrate SortingHat database if schema version < 0.8 ...",
```
sduenas, line 194: I think you should move this under the `if` statement, so it's only shown when the migration is taking place.

sduenas (changes requested):
The PR looks ok. However, you should update the README file to explain how the migration for this new command works - there's no need to call the utils script anymore.

sduenas (approved):
LGTM

PR #737, opened 2023-02-09
[client] Allow not to verify SSL certificates

jjmerchante:
This PR allows to perform SSL requests to a domain without verifying its certificate.

This PR changes the class for sgqlc endpoint. It now uses RequestsEndpoint because is easier to define a session with the verify option to False. With HTTPEndpoint it should be necessary to overwrite the urlopen function with a custom context with SSL disabled.

Update sgqlc package to lastest version.


sduenas (approved):
LGTM

PR #744, opened 2023-02-27
[ui] Set domain as top domain

### LINKED ISSUE #729: Feature request: add the posibility to mark is_top_domain
For each domain, there is a field named `is_top_domain` that is useful to avoid the insertion of sub-domains which belong to the same organization (i.e  eu.example.com, us.example.com) (more info [here](https://github.com/chaoss/grimoirelab-sortinghat/commit/ad9903a26a1c92f951431cb27760d7a4e7cdb38b)), 

When using "Add organization"  in the frontend, there is no possibility to mark that option. 

image

Would be useful to have that option to have better management over each domain and improve affiliation.

evamillan:
Adds the option to mark domains as top domains on the Organization's modal.
Fixes #729.

### DIFF
--- a/releases/unreleased/set-top-domain-from-ui.yml
+++ b/releases/unreleased/set-top-domain-from-ui.yml
@@ -1,0 +1,8 @@
+---
+title: Set top domain from UI
+category: added
+author: Eva Millán <evamillan@bitergia.com>
+issue: 729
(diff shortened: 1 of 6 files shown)


sduenas (approved):
LGTM

PR #755, opened 2023-04-05
Include multi-tenancy in SortingHat

jjmerchante:
This PR allows hosting multiple instances with a single service having each instance's data isolated in different databases.

To configure it you need to:
- Set `MULTI_TENANT` setting to `True`.
- Define the databases in a JSON file, you can use the default file provided at sortinghat/config/tenants.json
```
{
  "tenants": ["sh1", "sh2"]
}
```
- Run the configuration script and run SortintgHat
- Assign users to tenants with `set-user-tenant` command.
```
sortinghat-admin set-user-tenant admin localhost:8001 sh1
sortinghat-admin set-user-tenant admin localhost:8002 sh2
```

- Run a nginx to host multiple instances of SortingHat. `docker-compose.yml` and `nginx.conf` example files:
```
version: '2.2'
services:
    nginx:
      image: nginx:latest
      volumes:
        - ./nginx.conf:/etc/nginx/conf.d/default.conf
      network_mode: "host"
```
```
server {
  listen 8001;

  server_name localhost 127.0.0.1;

  location / {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $http_host;
     proxy_redirect off;
     proxy_pass http://127.0.0.1:8000;
  }
}

server {
  listen 8002;

  server_name localhost 127.0.0.1;

  location / {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header Host $http_host;
     proxy_redirect off;
     proxy_pass http://127.0.0.1:8000;
  }
}
```
- Run 2 instances of the UI (it can be deployed serving static files with NGINX):
```
yarn serve --api_url=http://localhost:8001/api/
```
```
yarn serve --api_url=http://localhost:8002/api/
```
In this case, if you login into both instances with the user admin, the changes in one of them won't be shown in the other and the other way around.


sduenas (changes requested):
In overall, it looks good but we should add some tests to check if the multitenancy works. Problably just something like to have two tenants, write individuals to them, and check data is correctly inserted.

Also, we should also check that the tenant name is correctly stored on the logs entries on the database.

sortinghat/config/settings.py
```diff
@@ -15,6 +15,8 @@
 
 import os
 
+from sortinghat.config.tenants import TENANT_NAMES
```
sduenas, line 18: I think it'll be better if we use a json file or similar to get the list of tenants. Also, I would define a new env variable named `MULTI_TENANT_LIST_PATH` so the users can define where the list of tenants are. By default it'll point to `BASE_DIR + "tenants.json"` (or similar). If the variable is define and `MULTI_TENANT` is `True`, multi tenancy will be activated.

sduenas (changes requested):

sduenas (approved):
LGTM

PR #756, opened 2023-04-05
[ui] Remove groups table

evamillan:
This PR removes the groups table from the dashboard and makes some small style changes to make the remaining tables more responsive.

### DIFF
--- a/releases/unreleased/remove-groups-table-from-ui.yml
+++ b/releases/unreleased/remove-groups-table-from-ui.yml
@@ -1,0 +1,10 @@
+---
+title: Groups table removed from the UI
+category: removed
+author: Eva Millán <evamillan@bitergia.com>
+issue: null
+notes: |
+  Groups and organizations are very similar, and having both tables
+  in the dashboard can be confusing to users.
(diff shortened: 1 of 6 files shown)


sduenas (changes requested):
It looks good but, would you mind to add the changelog entry explaining we have removed the groups section fromo the ui and why?

evamillan:
@sduenas I added the changelog entry.

releases/unreleased/remove-groups-table-from-ui.yml
```diff
@@ -0,0 +1,10 @@
+---
+title: Remove groups table from UI
```
sduenas, line 2:
```suggestion
title: Groups table removed from the UI
```

The title should be reworded because it's for the end user and not the title of the commit.

sduenas, line 2:
```suggestion
title: Groups table removed from the UI
```

The title should be reworded because it's for the end user and not the title of the commit.

sduenas (approved):
LGTM

PR #757, opened 2023-04-12
[cmd] Create new accounts using commands

jjmerchante:
This PR allows the creation of new accounts using Django commands. To make a new account use the following:
```
Usage: sortinghat-admin create-user [OPTIONS]

  Create a new user given a username and password

Options:
  --username TEXT   Specifies the login for the user.
  --is-admin        Specifies if the user is superuser.
  --no-interactive  Run the command in no interactive mode.
  --help            Show this message and exit.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,21 @@ or affiliation. To start a worker run the command:
 $ sortinghatw --config sortinghat.config.settings
 ```
 
+## Create new accounts
+To create new accounts for SortingHat use the following command:
+
+```
+(.venv)$ sortinghat-admin create-user
+
+Usage: sortinghat-admin create-user [OPTIONS]
+
+  Create a new user given a username and password
+
+Options:
+  --username TEXT   Specifies the login for the user.
+  --is-admin        Specifies if the user is superuser.
+  --no-interactive  Run the command in no interactive mode.
+```
 
 ## Compatibility between versions
 SortingHat 0.7.x is no longer supported. Any database using this version will not work.
--- a/releases/unreleased/create-user-command.yml
+++ b/releases/unreleased/create-user-command.yml
@@ -1,0 +1,8 @@
+---
+title: Create account command
+category: added
+author: Jose Javier Merchante <jjmerchante@bitergia.com>
+issue: null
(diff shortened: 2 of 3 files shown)


sortinghat/core/management/commands/create_user.py
```diff
@@ -0,0 +1,54 @@
...
+    help = "Create new a new account in SortingHat"
+
+    def add_arguments(self, parser):
+        parser.add_argument("username", help="Account username")
+        parser.add_argument("password", help="Account password")
+        parser.add_argument("--admin", action="store_true", help="Account created is admin")
+
+    def handle(self, *args, **options):
+        extra_fields = {}
+        if 'admin' in options and options['admin']:
```
sduenas, line 43:
```suggestion
        if options.get('admin', False):
```

sortinghat/core/management/commands/create_user.py
```diff
@@ -0,0 +1,54 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = "Create new a new account in SortingHat"
+
+    def add_arguments(self, parser):
+        parser.add_argument("username", help="Account username")
+        parser.add_argument("password", help="Account password")
```
sduenas, line 38: I wonder if we can use the same parser that django-admin uses when a new superuser is created.

sortinghat/core/management/commands/create_user.py
```diff
@@ -0,0 +1,54 @@
...
+import django.db
+from django.contrib.auth import get_user_model
+from django.core.management import BaseCommand
+
+User = get_user_model()
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
sduenas, line 33: I think this will be better to integrate it with the `sortinghat-admin` command, as a subcommand of it.

sduenas (changes requested):
The command should be integrated within `sortinghat-admin` command.

sduenas (changes requested):
LGTM, but please, add a changelog file

PR #758, opened 2023-04-17
[ui] Improve teams UI

evamillan:
This PR moves the full list of teams from a modal to the expanded organization and allows individuals to be dragged and dropped into a team to affiliate them and viceversa.


sduenas (changes requested):
I found a pair of bugs:

- When an individual is added to a team, it should be added to the organization too. I think this is something the API does.
no_org

- When an individual is added to a subteam (in the example, Bitergia->Engineering->devops), the team is hidden.
no_subteam

sduenas (approved):
LGTM

PR #759, opened 2023-04-20
Set node version to 16.x

sduenas:
Pinning node to 16.x ensures GitHub Actions is not going to use node 18 or later.


jjmerchante:
LGTM

PR #761, opened 2023-04-24
Set NodeJS version to 16.x on Docker workflow

sduenas:
To build the UI we need NodeJS 16. The current version of the workflow is using 18.x. This commit forces to use 16.x.


jjmerchante:
LGTM

PR #763, opened 2023-04-27
[ci] Do not checkout repository in build action

jjmerchante:
This PR disables the checkout step in the build action to include the JS static files built in the action.


sduenas (approved):
LGTM

PR #765, opened 2023-04-28
[tenants] Fix tenant selection in job context

jjmerchante:
This PR fixes an error that was caused when the context of a Job wasn't defined as keyword argument and the tenant was not selected correctly.


sduenas (approved):
LGTM

PR #766, opened 2023-04-28
[ui] Organizations view

evamillan:
This PR adds a page for each organization at `/organization/<organization name>`. This view allows users to see, update or delete the organization's teams, members and domains.
It also makes some of the properties of the individuals table optional in order to use a simpler version of it in this view.


sduenas (approved):
LGTM

PR #767, opened 2023-05-04
[tenant] Use multi-tenancy with headers instead of host

jjmerchante:
Multi-tenancy is now selected using the `sortinghat-tenant` header instead of the host. NGINX should include that header when acting as a proxy of the SortingHat server, and other services should add that header when requests are sent to the server.


sduenas (approved):
LGTM

PR #775, opened 2023-05-10
[jobs] Sort jobs from newest to oldest

evamillan:
Sorts the list of jobs from newest to oldest to show running jobs first.


sduenas (approved):
LGTM

PR #781, opened 2023-05-16
[uwsgi] Set UWSGI workers and threads in env variables

jjmerchante:
Define default UWSGI workers to 1 and UWSGI threads to 4 to run multiple requests in a single SortingHat instance.


sduenas (approved):
LGTM

sduenas (changes requested):
Would you mind to add the change log entry explaining that there are two new env variables that can be used to configure the workers and threads of the server?

sduenas (approved):
LGTM

PR #782, opened 2023-05-16
[unify] Fix unify job performance

jjmerchante:
This PR updates the way unify job fetches all the individuals from the database. Instead of retrieving identities from each individual, fetch all the identities at once.

### DIFF
--- a/sortinghat/core/recommendations/matching.py
+++ b/sortinghat/core/recommendations/matching.py
@@ -23,6 +23,7 @@
 
 import logging
 import pandas
+import numpy
 
 from collections import defaultdict
 from django.forms.models import model_to_dict
@@ -163,6 +164,7 @@ def _find_matches(set_x, set_y, criteria, exclude, verbose):
     def _to_df(data_set):
         """Convert identities set into a Pandas `Dataframe` object"""
         df = pandas.DataFrame(data_set)
+        df.replace('', numpy.nan, inplace=True)
         return df.sort_values(['individual'])
 
     def _apply_recommender_exclusion_list(df):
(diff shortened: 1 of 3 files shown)


sortinghat/core/recommendations/matching.py
```diff
@@ -94,13 +94,20 @@ def _get_identities(uuid):
...
+            input_set.update(identities)
+    else:
+        identities = Identity.objects.select_related('individual').all()
         input_set.update(identities)
+        for identity in identities:
+            if identity.individual.mk in aliases:
+                aliases[identity.individual.mk].append(identity.uuid)
+            else:
+                aliases[identity.individual.mk] = [identity.uuid]
+        source_uuids = aliases.keys()
```
sduenas, line 110: What about using `defaultdict`? https://docs.python.org/3/library/collections.html#defaultdict-examples

tests/test_jobs.py
```diff
@@ -594,8 +594,8 @@ def test_recommend_matches_all_individuals(self):
         }
 
         recommendations_expected = {
-            self.john_smith.individual.mk: [self.jsmith.individual.mk, self.js_alt.individual.mk],
-            self.jsmith.individual.mk: [self.js_alt.individual.mk],
+            self.js_alt.individual.mk: [self.jsmith.individual.mk, self.john_smith.individual.mk],
```
sduenas, line 597: Why does this test require the change?

jjmerchante: The order of the recommendations changed...

sduenas (changes requested):
LGTM but please, add a changelog file with something like:

```
Performance improved for recommendations and merging jobs

The performance of the matching and merging algorithms that
are used on these jobs have been considerably improved.
These jobs will be 4 times faster than on the previous version.
```

PR #783, opened 2023-05-16
[matching] Improve performance of matching recommendations

sduenas:


sduenas:
This change is already on #782

PR #784, opened 2023-05-17
[poetry] Include google-auth as direct dependency

jjmerchante:
This PR includes `google-auth` as a direct dependency because it was using an old version and the release failed.


sduenas (approved):
LGTM

PR #788, opened 2023-05-18
[ui] Show hiden buttons when the mouse is over the table row

### LINKED ISSUE #787: Improve buttons that appear when mouse is over them
The buttons to lock an individual or mark it as a bot only are only visible when the mouse is over the individual's name, which may make it hard to find them. They should appear when the mouse is over a wider area, like the entire table row.

evamillan:
Makes the buttons to lock an individual or mark it as bot visible when the mouse is over the table row instead of just the name.
Fixes #787.

### DIFF
--- a/releases/unreleased/buttons-on-hover.yml
+++ b/releases/unreleased/buttons-on-hover.yml
@@ -1,0 +1,9 @@
+---
+title: Show hidden buttons when the mouse is over the table row
+category: fixed
+author: Eva Millan <evamillan@bitergia.com>
+issue: 787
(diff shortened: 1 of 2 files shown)


sduenas (changes requested):
There must be an error because both buttons don't work correctly when clicking on them. They remain active when clicking several times on them. See the next video:

test_hoover

evamillan:
@sduenas I updated the PR, it should be fixed now.

sduenas (approved):
LGTM

PR #791, opened 2023-05-23
[ui] Fix enrollment in expanded individual

evamillan:
This PR fixes the issue with adding an organization to an individual using the "+ ADD" button on the table expanded item.


releases/unreleased/add-org-to-individual.yml
```diff
@@ -0,0 +1,8 @@
+---
+title: Add an organization to an individual
```
sduenas, line 2:
```suggestion
title: ADD button doesn't affiliate individuals to organizations
```

sduenas (approved):
LGTM

PR #796, opened 2023-06-20
[ui] Fix filter by enrollment when clicking several times

evamillan:
Fixes the issue with the individuals table showing the unfiltered list instead of the members of a team or organization when the button is clicked several times.


sduenas (approved):
LGTM

PR #803, opened 2023-07-21
Switch around BaseError.message and msg

s-t-e-v-e-n-k:
While investigating error messages when attempting to build this, the non-interplotated messages were leaking into the API. Further more, message being non-interplotated, and msg being interplotated felt backwards to me, so I've switched them around, and I've also hidden msg to _msg to make it a little clear that it's an internal detail only.


sduenas (approved):
LGTM

PR #804, opened 2023-07-24
Unify scheduled tasks

evamillan:
This PR removes the `ImportIdentitiesTask` model and all its related methods in order to avoid duplicate code. The tasks to import identities can now be scheduled using the generic `scheduleTask` GraphQL mutation, eg.:

```
mutation exampleImportTask {
  scheduleTask(
    job: "import_identities",
    interval: 100,
    params: "{\"backend_name\": \"mailmap\", \"url\": \"https://test.com/.mailmap\"}"
  ) {
    task {
      id
    }
  }
}
```

The tasks are created, updated, deleted and listed using the same methods as the `affiliate` and `unify` tasks. The GraphQL calls on the UI are updated but there is no visible change for the end user.


sduenas (approved):
LGTM

PR #806, opened 2023-08-01
[affiliation] Fix subdomain affiliation when it's not top domain

### LINKED ISSUE #805: Affiliate job doesn't support top domains
The `affiliate` job matches the organizations subdomains regardless of whether they are marked as "top" domains or not. Eg., it affiliates the individuals with `@users.noreply.github.com` emails to GitHub when it should only match `@github.com`.

evamillan:
This PR prevents the affiliation jobs from matching subdomains if the domain is not marked as `top_domain`.
Fixes #805.


sduenas (approved):
LGTM

PR #807, opened 2023-08-21
UI to manage settings

evamillan:
This PR adds a `Settings` section where users can manage the `affiliate` and `unify` tasks. `Import identities` and `jobs` are also moved to this section .

### DIFF
--- a/ui/src/components/TasksTable.vue
+++ b/ui/src/components/TasksTable.vue
@@ -1,7 +1,7 @@
 <template>
-  <v-container class="section jobs mt-6 pa-0">
+  <section class="section">
     <header class="header">
-      <h1 class="title">Import identities</h1>
+      <h1 class="title">Synchronization</h1>
       <v-menu offset-y>
         <template v-slot:activator="{ on }">
           <v-btn color="primary" depressed small v-on="on">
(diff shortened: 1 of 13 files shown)


ui/src/components/TasksTable.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-  <v-container class="section jobs mt-6 pa-0">
+  <section class="section">
     <header class="header">
-      <h1 class="title">Import identities</h1>
+      <h1 class="title">Sync data</h1>
```
sduenas, line 4: I would rename it to "Synchronization"

ui/src/views/SettingsLayout.vue
```diff
@@ -0,0 +1,59 @@
...
+              <v-list-item-title>General</v-list-item-title>
+            </v-list-item-content>
+          </v-list-item>
+          <v-list-item
+            :to="{ name: 'SettingsSync' }"
+            active-class="primary--text white"
+            link
+          >
+            <v-list-item-content>
+              <v-list-item-title>Sync data</v-list-item-title>
```
sduenas, line 26: Same as above

ui/src/views/SettingsLayout.vue
```diff
@@ -0,0 +1,59 @@
...
+            </v-list-item-content>
+          </v-list-item>
+          <v-list-item
+            :to="{ name: 'SettingsJobs' }"
+            active-class="primary--text white"
+            link
+          >
+            <v-list-item-content>
+              <v-list-item-title>Job log</v-list-item-title>
```
sduenas, line 35: This should just be Jobs

sduenas (changes requested):
In general, it looks good to me. Just minor modifications.

evamillan:
@sduenas I made the changes you requested and I also added [this commit](https://github.com/chaoss/grimoirelab-sortinghat/pull/807/commits/a2e19b0091ca7b048d86f8b30b20cdbecde6a0f9) to delete the task from the database if the job couldn't be created because there was no Redis connection. This way the configuration won't appear as active on the UI.

sortinghat/core/jobs.py
```diff
@@ -779,7 +780,11 @@ def create_scheduled_task(ctx, job, interval, params):
     if not params:
         params = dict()
 
-    schedule_task(ctx, job_fn, task, **params)
+    try:
+        schedule_task(ctx, job_fn, task, **params)
+    except redis.exceptions.ConnectionError as e:
+        delete_scheduled_task(ctx, task.id)
```
sduenas, line 786: I think this is ok but it contradicts the statement on line where we call to `add_scheduled_tasks`. This one is a `db` operation while `delete_scheduled_task` is a api call. For our mental health, I think you should create a `add_schedule_task` in api and replace that call with it.

I think you can remove the `delete` call here and open a new PR where you create that new api function to fix this problem. Do you mind to do that?

sduenas (changes requested):
Thanks for the changes. They look find. However, I think your last commit should go in a different PR. I give more info in the inline comments.

sortinghat/core/jobs.py
evamillan: No problem, I removed the last commit and I'll open a PR with that.

PR #810, opened 2023-08-28
Delete scheduled tasks with no job due to Redis connection error

evamillan:
This PR adds an API function to create scheduled tasks and fixes the issue with the tasks that have no enqueued job due to a Redis connection error when they are created.

### DIFF
--- a/releases/unreleased/delete-task-on-error.yml
+++ b/releases/unreleased/delete-task-on-error.yml
@@ -1,0 +1,11 @@
+---
+title: Remove tasks that fail to be scheduled
+category: fixed
+author: Eva Millan <evamillan@bitergia.com>
+issue: null
+notes: >
+  When there was an issue with the Redis connection when
+  a task was created, the task was added to the database
+  but there was not scheduled job linked to it. Tasks are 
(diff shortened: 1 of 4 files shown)


sortinghat/core/jobs.py
```diff
@@ -774,12 +775,16 @@ def create_scheduled_task(ctx, job, interval, params):
 
     trxl = TransactionsLog.open('create_scheduled_task', ctx)
```
sduenas, line 776: The transaction object is not needed anymore.

releases/unreleased/delete-task-on-error.yml
```diff
@@ -0,0 +1,11 @@
+---
+title: Remove tasks that fail to be scheduled
```
sduenas, line 2: You should add another changelog file explaining we are adding the `add_scheduled_task` to the api

sduenas (changes requested):
Please check my comments

sduenas (approved):
LGTM

sduenas:
@evamillan I messed it up with your branch updating it. Can you revert the changes, please?

PR #814, opened 2023-08-31
Strict criteria for merge recommendations

### LINKED ISSUE #812: Recommendations with extrict critera
The old SortingHat (0.7) had a feature that allowed to unify identities only when the fields used to evaluate a match satisfy some rules.

For example, to use the email field, this has to be a valid email address (example@example.com). For the name field, only those with a first name and one or more last names (e.g. John Smith, John Smith Smith) were selected.

The latest version of SortingHat doesn't have this feature. It should be something that is available by default, but it can be deactivated when needed.

evamillan:
This PR adds the `strict` parameter to the merge recommendations, which evaluates if the individual's names and emails are valid when looking for matches. This is enabled by default.
Fixes #812.

### DIFF
--- a/releases/unreleased/strict-criteria.yml
+++ b/releases/unreleased/strict-criteria.yml
@@ -1,0 +1,10 @@
+---
+title: Strict criteria for merge recommendations
+category: added
+author: Eva Millan <evamillan@bitergia.com>
+issue: 812
+notes: >
+  The merge recommendations filter out invalid email adresses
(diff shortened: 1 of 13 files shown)


evamillan:
I updated the PR with the tests from `test_jobs`.

evamillan:
I also added the UI to configure the parameter.

jjmerchante:
LGTM

PR #827, opened 2023-10-26
Cache individuals queries to avoid reloading the table

### LINKED ISSUE #821: Avoid reloading table search when individual's data is changed
When an individual's information is changed on the table, eg. when it is merged, the current search on the table is reloaded. This makes the process very slow when there is a huge number of identities.

evamillan:
This PR sets the fetch policy of the individuals queries to `cache-first`. When a mutation changes the data of a cached query, Apollo uses the returned data to update the cache automatically, so no refetching is needed. Apollo can't update the queries if items are added or removed from a list, so after a merge, unmerge, create or delete, or if the query was sorted by `lastModified` or was filtered by an organization that was added or removed from the individual, the queries need to be refetched.
Fixes #821

### DIFF
--- a/releases/unreleased/cache-individuals-table-data.yml
+++ b/releases/unreleased/cache-individuals-table-data.yml
@@ -1,0 +1,14 @@
+---
+title: Cache individuals table data
+category: added
+author: Eva Millán <evamillan@bitergia.com>
+issue: 821
+notes: >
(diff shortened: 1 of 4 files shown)


jjmerchante:
LGTM

sduenas (approved):
LGTM

PR #833, opened 2023-11-13
[ui] Fix uninitialized workspace

evamillan:
This PR fixes the issue with the individual's page not loading. It initializes the workspace as an empty array if there is not one saved on the local storage. The individual's view also checks if the workspace exists before checking for the individual to prevent the view from not loading if there is no workspace.


sduenas (approved):
LGTM

PR #834, opened 2023-11-16
Improve performance of recommender job

jjmerchante:
This commit improves the performance of the recommender job:
- It now only groups individuals that are directly related. This means that if A = {B} and B = {A, C}, it won't create a recommendation between A and C.
- The recommendations are now created between individuals instead of identities. This results in fewer recommendations but is more useful if you only need to merge individuals. If you need recommendations between identities, set verbose to True.
- It no longer queries the database to check if a recommendation is created; it follows the style "Easier to Ask Forgiveness Than Permission" (EAFP).
- The recommendations internally now return the mk of an identity, avoiding unnecessary database queries.


sortinghat/core/jobs.py
```diff
@@ -443,23 +454,34 @@ def unify(ctx, criteria, source_uuids=None, target_uuids=None, exclude=True, str
...
+                    dfs(neighbor, current_set)
+
+        result = []
+        for key in recs:
+            if key not in visited:
+                current_set = set()
+                dfs(key, current_set)
+                if len(current_set) > 1:
+                    result.append(current_set)
+
```
sduenas, line 483: I think this is a bit complicated. I don't like that recursive function and I don't like it's using global variables. A better approach for DFS is to think that you will need to empty a stack of nodes.
Every time you discover a new node, you add it to the stack, on the next iteration you get the first element from the stack. You do that until the stack is empty. When it's empty, you have a path and you start processing a new path using a new empty stack. Because there can be cycles, you should still use the visited set to avoid doing the work twice.

sduenas (changes requested):
In overall it looks ok but there are some changes needed. Please address them.

sortinghat/core/jobs.py
jjmerchante: I updated the function, let me know if the new algorithm is clearer

sduenas (approved):
LGTM

PR #836, opened 2023-11-30
[affiliation] Improve performance of affiliation jobs

sduenas:
The current algorithm was running unnecessary calls to 'find_individual_by_uuid' when the generic affiliation job was called. In other words, when the list of individuals to affiliate was all the individuals in the database.
To improve it, we have followed the same approach we did with the matching recommendations. We get the full list of individuals with a direct query using Django's API.

### DIFF
--- a/sortinghat/core/jobs.py
+++ b/sortinghat/core/jobs.py
@@ -151,7 +151,6 @@ def recommend_affiliations(ctx, uuids=None, last_modified=MIN_PERIOD_DATE):
 
     if not uuids:
         logger.info(f"Running job {job.id} 'recommend affiliations'; uuids='all'; ...")
-        uuids = Individual.objects.filter(last_modified__gte=last_modified).values_list('mk', flat=True).iterator()
     else:
         logger.info(f"Running job {job.id} 'recommend affiliations'; uuids={uuids}; ...")
         uuids = iter(uuids)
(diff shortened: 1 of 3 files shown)


sortinghat/core/jobs.py
```diff
@@ -168,25 +167,24 @@ def recommend_affiliations(ctx, uuids=None, last_modified=MIN_PERIOD_DATE):
...
-    for chunk in _iter_split(uuids, size=MAX_CHUNK_SIZE):
-        for rec in engine.recommend('affiliation', chunk):
-            results[rec.key] = rec.options
+    for rec in engine.recommend('affiliation', uuids, last_modified):
+        results[rec.key] = rec.options
+        try:
+            individual = find_individual_by_uuid(rec.key)
+        except NotFoundError:
+            logger.warning(f"Job {job.id} 'Individual {rec.key} not found'")
+            continue
```
jjmerchante, line 179: I think we don't need this part. We have the `individual.mk` in `rec.mk`, and we can create the `AffiliationRecommendation` object using `rec.mk`, as we do in `recommend_matches()` with the following code:
```
try:
    with transaction.atomic():
        AffiliationRecommendation.objects.create(individual_id=rec.mk, organization=org)
except IntegrityError:
    pass
```

We can also use `rec.key` because `recommend_affiliations()` yields the same value for both.

jjmerchante (changes requested):

sortinghat/core/jobs.py
sduenas: I think we can only use `rec.mk` because if the list of `uuids` is not empty it will mean to create recommendations for those uuids only. In that case `key` is the `uuid` and not the `mk`

jjmerchante:
LGTM

PR #839, opened 2023-12-11
[ui] Move date picker to the side of the input

### LINKED ISSUE #838: Text field not visible when editing an affiliation
When clicking on edit a date of an affiliation from the identity view, it always shows the calendar to chose a date. This is ok but if the user knows the date, it's very hard to set it to this date with the text field because it isn't visible.

See the next capture to reproduce the problem:

date

evamillan:
This PR changes the position of the affiliation date calendar to open on the right side of the date input to avoid covering it.
Fixes #838

image


sduenas (approved):
LGTM

PR #842, opened 2023-12-13
[ui] Improve organization selector loading times

evamillan:
This PR changes the `OrganizationSelector` component used to affiliate individuals to only make a request after 500ms after the last key press. The query is also changed to a cached one that only returns the organizations names instead of all their data to make it faster.


sduenas (approved):
LGTM

PR #843, opened 2023-12-15
Pin urllib3 to ^1.26

jjmerchante:
This change is to keep packages aligned with the same version. Perceval requires urllib3 v1 because there is a bug with dulwich and urllib3.


sduenas (approved):
LGTM

PR #844, opened 2023-12-21
Option to match individuals with the same source

jjmerchante:
This PR includes a new option to unify and recommend identities that share the defined criteria (name, username, email) and the source.

The option is called `match_source` when creating the job.


sduenas (changes requested):
The PR looks good but I would like that this matching only works for sources that we know they are the same. In other words, services: github, slack, gitlab, etc.

sduenas (approved):
LGTM

PR #852, opened 2024-01-25
[ui] Fix API URLs to login and change password

### LINKED ISSUE #851: ChangePassword calls /password_change instead of /identities/password_change
Hi everyone,

while playing around with sortinghat, I noticed that the [`ChangePassword` view](https://github.com/chaoss/grimoirelab-sortinghat/blob/master/ui/src/views/ChangePassword.vue) calls `/password_change` instead of `/identities/password_change` which leads to a 404.

Without investigating further, I assume that the `url` is not computed correctly [here](https://github.com/chaoss/grimoirelab-sortinghat/blob/master/ui/src/views/ChangePassword.vue#L86). On the other hand, [here](https://github.com/chaoss/grimoirelab-sortinghat/blob/master/ui/src/store/index.js#L30) it is computed correctly for the `login` action.

I was able to fix this by adding a rewrite rule to nginx, but it's probably better to handle this directly within sortinghat. If you agree, I'm happy to set up a dev environment in order to propose a PR regarding this issue.

Best regards,

Eduard

## How to reproduce

1. Use the default [docker-compose.yml](https://github.com/chaoss/grimoirelab/blob/master/docker-compose/docker-compose.yml) file and start a local stack
2. Go to http://localhost:8000/identities/change-password and try to change the password
3. The browser console throws
```Failed to load resource: the server responded with a status of 404 (Not Found)      :8000/password_change/:1```
4. The password hasn't been changed

## Expected behaviour

The password is changed after submitting http://localhost:8000/identities/change-password

evamillan:
This PR changes the login and change password URLs to use the public path as a base if no API URL was specified, eg. `/identities/` in production.
Fixes #851.

### DIFF
--- a/releases/unreleased/fix-api-url-to-change-password.yml
+++ b/releases/unreleased/fix-api-url-to-change-password.yml
@@ -1,0 +1,8 @@
+---
+title: Use correct base URL for login and change password API calls
+category: fixed
+author: Eva Millán <evamillan@bitergia.com>
+issue: 851
+notes: >
+  The URLs called to login and change password now use the public
+  path found in vue.config.js if no API URL is specified.
--- a/ui/src/store/index.js
+++ b/ui/src/store/index.js
@@ -25,13 +25,14 @@ export default new Vuex.Store({
   actions: {
     async login({ commit }, { username, password }) {
       const csrftoken = Cookies.get("csrftoken");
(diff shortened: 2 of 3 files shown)


sduenas (approved):
LGTM

PR #853, opened 2024-01-30
Fix AUTHENTICATION_REQUIRED setting

jjmerchante:
When the AUTHENTICATION_REQUIRED setting is set to False, the authorization decorators allow any user to take any action.


sduenas (approved):
LGTM

PR #854, opened 2024-02-01
Fix missing 'django_sessions' table when using tenants

evamillan:
Adds the `sessions` app to the tenant database router to create the `django_sessions` table on the default database to fix the `Table 'django_session' doesn't exist` error. For existing databases, run the following commands to create the table:
  ```
  django-admin migrate --fake sessions zero
  django-admin migrate
  ```


jjmerchante (approved):
LGTM

PR #858, opened 2024-02-16
Add organization aliases

### LINKED ISSUE #857: Organization aliases
Sometimes we have been in the situation that an organization can have different names.  The proposal is that the organizations can have aliases.

This new feature should:

- Add aliases
- Remove aliases
- When one organization is merged into another one, its name will become an alias
- When searching for organizations, if the alias is given, the search should return the organization that has that alias
- If a name exists as an alias, no organization can be created with that name.

evamillan:
This PR adds aliases for the organizations.
- Searching for an organization using one of its aliases returns the organization.
- When an organization is merged into another, its name becomes an alias of the target organization.
- If a name exists as an alias, no organization can be created with that name and viceversa.

Fixes #857.


sortinghat/core/db.py
```diff
@@ -1018,6 +1047,45 @@ def move_domain(trxl, domain, organization):
...
+    :raises ValueError: raised when the alias is already assigned to
+    the organization
+    """
+    op_args = {
+        'alias': alias.alias,
+        'organization': organization.name
+    }
+
+    if alias.organization.name == organization.name:
+        msg = f"Domain '{alias.alias}' is already assigned to '{organization.name}'"
```
sduenas, line 1072: It should be `Alias` instead of `Domain`.

sortinghat/core/api.py
```diff
@@ -1330,6 +1430,13 @@ def _move_domains(trxl, from_org, to_org):
         for domain in domains:
             move_domain(trxl, domain, to_org)
 
+    def _move_aliases(trxl, from_org, to_org):
+        """Move the domains from the origin organization to the target"""
```
sduenas, line 1434: Reference to alias instead of domains.

tests/test_db.py
```diff
@@ -373,6 +389,56 @@ def test_domain_name_int(self):
...
+class TestFindAlias(TestCase):
+    """Unit tests for find_alias"""
+
+    def setUp(self):
+        """Load initial dataset"""
+
+        org = Organization.add_root(name='Example')
+        Alias.objects.create(alias='Example Inc.', organization=org)
+
+    def test_find_domain(self):
```
sduenas, line 401: Alias instead of domain

tests/test_db.py
```diff
@@ -373,6 +389,56 @@ def test_domain_name_int(self):
...
+
+        alias = db.find_alias('Example Inc.')
+
+        # Tests
+        self.assertIsInstance(alias, Alias)
+        self.assertEqual(alias.organization.name, 'Example')
+        self.assertEqual(alias.alias, 'Example Inc.')
+
+    def test_alias_not_found(self):
+        """Test whether it raises an exception when the domain is not found"""
```
sduenas, line 412: Domain again

tests/test_db.py
```diff
@@ -3189,3 +3259,209 @@ def test_operations(self):
...
+
+        # Tests
+        with self.assertRaises(ObjectDoesNotExist):
+            Alias.objects.get(alias='Example Inc.')
+
+        org.refresh_from_db()
+        self.assertEqual(len(org.aliases.all()), 1)
+
+    def test_operations(self):
+        """Check if the right operations are created when deleting a domain"""
```
sduenas, line 3443: Domain again.

sduenas (changes requested):
Just minor changes regarding references to domains instead to aliases.

sduenas (changes requested):
Testing this PR from the user interface I got the next error:
imagen

I got the error creating an organization with two aliases and trying to delete it. With one alias there's no error. Same error is returned when organizations with more than one alias are merged.

evamillan:
@sduenas I removed the references to domain, fixed the error and added some test cases for organizations with more than one alias.

sduenas (approved):
LGTM

PR #860, opened 2024-02-21
Add support for Node.js 18-20 and drop 14-16

evamillan:
This PR drops the support for unmantained versions 14 to 16 of Node and updates the dependencies to support versions 18 to 20. It also fixes the rest of the warnings raised by the GitHub test actions.


sduenas (approved):
LGTM

PR #861, opened 2024-02-28
[jobs] Define dedicated queues for tenants

sduenas:
So far, jobs could only run in the default RQ queue. With this change, tenant jobs can run in dedicated queues when needed.

To implement this feature, it was neccesary to update the creation of the databases and queues in the
configuration file. Also, a new auto-generated config variable named TENANTS_DEDICATED_QUEUES has been set to determine whether a tenant runs its jobs in a
dedicated queue or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,13 @@ or affiliation. To start a worker run the command:
 $ sortinghatw --config sortinghat.config.settings
 ```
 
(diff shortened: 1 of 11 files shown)


jjmerchante:
LGTM

jjmerchante (approved):

PR #862, opened 2024-02-29
[job] Fix recommend matches job

jjmerchante:
This PR fixes the recommend matches job when executed from the settings/job that was causing it not to recommend and match.


sduenas (approved):
LGTM

PR #863, opened 2024-03-01
[docker] Set build image with node 20.x

sduenas:
We missed to update the version of node when the docker image is created.


jjmerchante:
LGTM

PR #871, opened 2024-03-18
[ui] Migrate to Vue 3

evamillan:
This PR migrates SortingHat to Vue 3. Some notes:
- Not all the dependencies are updated, only those required to work with Vue 3. Vue-cli is not longer maintained and they recommend Vite for all new projects. Storybook has released new versions but they remove Storyshots, so we'd have to find another way of testing snapshots.
- The Vuetify tree component is still in beta and has many bugs, so it is switched to another component for the moment. This means the teams only have 2 levels of hierarchy on the UI and can't display unlimited teams inside teams.


sduenas (approved):
LGTM

PR #873, opened 2024-03-22
[ui] Fix affiliate and unify cache

evamillan:
This PR adds the `no-cache` policy to the scheduled tasks query in order to refetch it when the tasks are created or deleted.
It also removes the Apollo Link packages since their functionality is now included in the core Apollo client.


sduenas (approved):
LGTM

PR #879, opened 2024-04-12
Upgrade to Django 4.2

jjmerchante:
This commit upgrades the Django version from 3.2 to 4.2. Support for Django 3.2 concluded at April 1, 2024. This update ensures compatibility with the latest features and an extended support.

Some explanation about the changes:
- The tests were importing `testing.py` and `testing_tenant.py` because they start with `test*.py` and the tests were failing for another reason that I could not identify. Renamed to `config_testing.py` and `config_testing_tenant.py`
- The pagination error when `num_pages` is negative has changed.
- `USE_L10N` was removed and is not needed anymore
- `django.utils.timezone.utc` is deprecated in version 4.1
- `check_permissions` is using `user.has_perms()` which requires an iterable


sduenas (approved):
LGTM

PR #884, opened 2024-04-18
Clarify recommededMerge

fioddor:


sduenas (changes requested):
Please, add a better [commit message](https://github.com/chaoss/grimoirelab/blob/master/CONTRIBUTING_WITH_CODE.md#how-to-write-good-commit-messages).

Also, use this method to create multi-line strings. With your patch, the string will have the blank spaces before the text. They way to fix it is:

```python
description = (
    "this is a multi-line"
    "string to include"
    "as description"
)
```

fioddor:
> Please, add a better [commit message](https://github.com/chaoss/grimoirelab/blob/master/CONTRIBUTING_WITH_CODE.md#how-to-write-good-commit-messages).

- [x] Separate the subject from the body with a blank line.
- [x] Limit the subject line to 50 characters.
- [x] Capitalize the subject line and (optionally) start the subject with a tag.
- [x] Do not end the subject line with a period.
- [x] Use the imperative mood in the subject line.
- [x] Wrap the body at 72 characters.
- [x] Use the body to explain what and why rather than how.

> Also, use this method to create multi-line strings. With your patch, the string will have the blank spaces before the text. They way to fix it is:
> 
> ```python
> description = (
>     "this is a multi-line"
>     "string to include"
>     "as description"
> )
> ```

:+1:

PR #892, opened 2024-06-11
Fix unify individuals with the same username and source

### LINKED ISSUE #875: Match individuals for the same source should only work for username
This feature - introduced by https://github.com/chaoss/grimoirelab-sortinghat/pull/844 - was only meant to be available when matching data by `username` and not by other criteria. It applies to all the criteria.

jjmerchante:
This PR resolves an issue where the "match source" option was being applied to emails and names. With this update, this option will now only apply to usernames. Additionally, users can use email and name simultaneously.

Fixes #875

### DIFF
--- a/config/settings/devel.py
+++ b/config/settings/devel.py
@@ -179,3 +179,5 @@ RQ_QUEUES = {
         'DB': 0
     }
 }
+
+MULTI_TENANT = os.environ.get('SORTINGHAT_MULTI_TENANT', 'False').lower() in ('true', '1')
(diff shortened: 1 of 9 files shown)


sduenas (changes requested):
It works great. Thanks for the PR.
I have a request though. Would you mind to add this line to the devel config? When testing everything the devel settings didn't work because that setting is not set.
```
MULTI_TENANT = os.environ.get('SORTINGHAT_MULTI_TENANT', 'False').lower() in ('true', '1')
```

sduenas (approved):
LGTM

PR #897, opened 2024-07-04
Add readonly and read/write permission groups

evamillan:
This PR adds the `readonly` and `user` permission groups. The `user` group has add, change, delete and view permissions for all `core` models except for jobs and task scheduling. The `readonly` group only grants view permissions.
The groups and their permissions are moved to `sortinghat/config/permission_groups.json`. The location of the file can be changed with the `SORTINGHAT_PERMISSION_GROUPS_LIST_PATH` environment variable.

The PR also adds a command to assign a user to a group:
```
$ sortinghat-admin set-user-group username group
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,15 @@ Options:
   --no-interactive  Run the command in no interactive mode.
 ```
 
+## Assign users to permission groups
+A user in a group automatically has the permissions granted to that group. To assign users to a permission group use the following command:
+```
+$ sortinghat-admin set-user-group username group
+```
+
+The list of groups can be customized using the configuration file `sortinghat/config/permission_groups.json`. You can use a different json file using the environment variable `SORTINGHAT_PERMISSION_GROUPS_LIST_PATH`.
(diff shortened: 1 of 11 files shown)


jjmerchante:
The only problem is `importidentitiestask` permission, it fails to import that permission from the JSON and later can't be used.
Permissions are used for database models and `importidentitiestask` is not a table in the database anymore, it was in an older version before it was renamed to ScheduledTask.
That permission is needed for `ImportIdentities` mutation (not used from the UI), but is available from the API.

One option could be adding new permission to `CustomPermissions` table, along with `execute_job`, and could be used similarly. Another option is to remove that permission and use the `execute_job` permission for the `ImportIdentities` mutation, which make sense because the `ImportIdentites` mutation is a job.

evamillan:
I think it makes sense to use `execute_job` too, I updated the PR to use it instead.

jjmerchante (approved):
LGTM

PR #903, opened 2024-07-26
[config] Include variable to read CORS allowed origins using regex

jjmerchante:
Include `CORS_ALLOWED_ORIGIN_REGEXES` in the SortingHat server configuration that contains a list of strings representing regexes that match Origins that are authorized to make cross-site HTTP requests.


sduenas (approved):
LGTM

PR #905, opened 2024-07-30
[ui] Use consistent date format for jobs

### LINKED ISSUE #901: Different date format in the Jobs
I executed it several times and the date formats were different

- `16/7/2024, 18:45:06`

image

- `7/17/2024, 6:56:09 PM`

image

It would be great if the date format were the same for every execution.

evamillan:
This PR makes all the job execution dates use the same `YYYY-MM-DD hh:mm` format. The format logic is moved to a plugin to avoid having a different method in each component. The date formatting is managed by the `dayjs` library, which was already installed as a peer dependency.

Fixes #901.

### DIFF
--- a/releases/unreleased/consistent-date-format.yml
+++ b/releases/unreleased/consistent-date-format.yml
@@ -1,0 +1,8 @@
+---
+title: Consistent date format
+category: fixed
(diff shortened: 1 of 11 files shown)


jjmerchante (approved):
LGTM

PR #908, opened 2024-08-06
Fix jobs missing from results

jjmerchante:
This PR fixes a bug that caused an empty list of job results when one of them was missing in Redis. This PR also updates the time the job results are kept to one week.

This also updates the worker entry point to receive parameters when a container runs.


sduenas (approved):
LGTM

sduenas (changes requested):
flake8 is giving some errors. Can you fix them?

sduenas (approved):

PR #911, opened 2024-08-22
Define permissions for users per tenant

jjmerchante:
This PR updates the permissions system to allow assigning permissions to users based on the tenant they are accessing. The command `sortinghat-admin set-user-group` has been renamed to `sortinghat-admin set-user-permissions`.


sduenas (approved):
LGTM

Issue #849, opened 2024-01-15
Read only user


sduenas:
Sometimes is good to have a user that can have access to see only the identities but not to modify certain things. It would be nice to have this kind of user (read only user).

sduenas:
The current solution doesn't take into account `multitenancy`, so users have the same permissions on all the tenants. The PR #911 fixes this problem but we still need to add a method for updating the permissions when migrating database versions.

PR #916, opened 2024-08-28
Migrate user permissions from previous release

jjmerchante:
The previous release included the users' permissions in the Group Django model. In the new release, they are included in the Tenant model. This code migrates the permissions for all the users.

Related to https://github.com/chaoss/grimoirelab-sortinghat/issues/849

### DIFF
--- a/releases/unreleased/users-permissions-migrated.yml
+++ b/releases/unreleased/users-permissions-migrated.yml
@@ -1,0 +1,9 @@
+---
+title: Users permissions migrated
+category: fixed
+author: Jose Javier Merchante <jjmerchante@bitergia.com>
+issue: 849
+notes: >
+  User permissions are stored in a separate table within
+  the database. These permissions will be automatically
+  migrated when running `sortinghat-admin upgrade`.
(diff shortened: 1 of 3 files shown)


sduenas (approved):
The code is ok but the problem that I see is the first time the user migrates. It won't be any user an any group.

jjmerchante:
If they are included in a tenant, they will have `readonly` permissions. I can change that to `user` permissions.

sduenas:
> If they are included in a tenant, they will have `readonly` permissions. I can change that to `user` permissions.

Is that true? I mean, they have those permissions but there's no entry in the database saying that they have those permissions. The table `auth_user_groups` is empty.

jjmerchante:
> Is that true? I mean, they have those permissions but there's no entry in the database saying that they have those permissions. The table `auth_user_groups` is empty.

I think you mean that I forgot to migrate the permissions when multitenancy is disabled. I have updated that. Additionally, I changed the default permissions to `user` when multi-tenant is enabled, to match what it was before the update.

sduenas (approved):
LGTM

sduenas:
> > Is that true? I mean, they have those permissions but there's no entry in the database saying that they have those permissions. The table `auth_user_groups` is empty.
> 
> I think you mean that I forgot to migrate the permissions when multitenancy is disabled. I have updated that. Additionally, I changed the default permissions to `user` when multi-tenant is enabled, to match what it was before the update.

Yes, that's what I meant.

PR #917, opened 2024-08-28
[ui] Merge existing organizations from the organizations edit view

### LINKED ISSUE #913: Merge existing organizations from the organizations edit view
SortingHat's UI allows to merge organizations dragging the alternative and dropping it onto the main one.
As result, the alternative organization becomes an alias of the main one.

This works well for short lists but we have instances with tens of thousands of organizations where drag-drop becomes impossible.

The UI also allows to open an organization and edit their aliases. This checks the existence of the new alias as an existing organization and if so, it pops up a dialogue informing about that.

The popup should allow to merge the organizations showing the following message (or something similar):

```
    Warning creating alias

    Organization '<NEW_ALIAS>' already exists in the registry.
    Click on Merge to turn into an alias of '<OPEN_ORGANIZATION>' or Cancel.
```

evamillan:
This PR adds the option on the organization view, when adding a new alias that is an existing organization, to merge them. It also refactors the modal's code to simplify it and reduce code duplication.
Fixes #913.

### DIFF
--- a/releases/unreleased/merge-organizations-when-adding-an-alias.yml
+++ b/releases/unreleased/merge-organizations-when-adding-an-alias.yml
@@ -1,0 +1,10 @@
+---
+title: Merge organizations when adding an alias
+category: added
+author: <evamillan@bitergia.com>
+issue: 913
+notes: >
+  When adding an alias on an organization's page, if the
+  new alias is an existing organization, users now have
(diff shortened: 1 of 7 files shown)


sduenas (approved):
LGTM. It works as expected.

PR #918, opened 2024-08-30
Close DB connection before job completion

jjmerchante:
RQ workers create a fork to run the function. The issue arises when, after completing the job, the MariaDB logs display the warning: `Aborted connection to db. Got an error reading communication packets`. This change ensures the database connection is closed before the fork ends, preventing the warning from appearing.


sduenas (changes requested):
LGTM but can we call the new class `SortingHatJob` or something like that. Then, explain in the docstring that it's a class to change the default behaviour of RQ Job, and describe the problem with the connection (as it's described now in the docstring).

sduenas (approved):
LGTM

PR #922, opened 2024-09-03
Updated the worker image to load custom settings.

jjmerchante:
The worker container was not loading the custom settings due to the environment variable was being replaced.


sduenas (approved):
LGTM

PR #924, opened 2024-09-10
[ui] Fix split identities only added to the workspace once

### LINKED ISSUE #919: Split identities only adds them to the workspace once
When an individuals is split into many others, they are shown in the workspace. They will be shown the first time only. If the user does it with other individuals or the same, nothing will shown in the workspace.

See the video to reproduce the error:

[sh_workspace.webm](https://github.com/user-attachments/assets/ce86cbf6-62d4-45ac-9de6-aa591976f49e)

evamillan:
This PR fixes the issue with split identities only being added to the workspace the fist time they are split.
Fixes #919.


jjmerchante (approved):
LGTM

PR #926, opened 2024-09-11
[ui] Remove main identity star icon

### LINKED ISSUE #920: Remove the yellow star (main identity)
The main identity of an individual has no practical effect for UI users.

The yellow star marking it in the UI calls the users' attention and makes them wonder if they should pick another identity as the main one (which cannot be done easily in the UI).

Only for advanced technical users might it make any sense, and even in such case, choosing a more subtle visual indicator (like a small dot) would prevent misleading the average users.

evamillan:
This PR removes the star icon that marks an identity as the main one and replaces it with a darker background for the uuid chip.
Fixes #920.

### DIFF
--- a/releases/unreleased/less-distracting-main-identity-indicator.yml
+++ b/releases/unreleased/less-distracting-main-identity-indicator.yml
@@ -1,0 +1,10 @@
+---
+title: Less distracting main identity indicator
+category: added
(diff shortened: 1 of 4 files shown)


jjmerchante (approved):
LGTM

PR #927, opened 2024-09-12
[ui] Redirect to the original URL after log in

### LINKED ISSUE #925: Redirect to the original URL after log in
When the user introduces an URL and they aren't log in, the system redirects to the log in page. After they are log in, the system sends the user to the main page and not to the one originally introduced by the user.

The system should redirect to the original URL after log in.

evamillan:
This PR adds a parameter to the /login route to save the URL that the user originally requested and redirect to it after authentication.
Fixes #925.


releases/unreleased/redirect-to-the-original-url-after-log-in.yml
```diff
@@ -0,0 +1,8 @@
+---
+title: Redirect to the original URL after log in
+category: added
+author: Eva Milán <evamillan@bitergia.com>
```
jjmerchante, line 4: LGTM, works as expected. Please check your author's surname, this will create a new entry in the AUTHORS file :smile: https://github.com/chaoss/grimoirelab-sortinghat/blob/master/AUTHORS

jjmerchante (changes requested):

jjmerchante (approved):
Thank you. LGTM

PR #928, opened 2024-09-16
Customizable trusted sources for username matching

jjmerchante:
Enable customization of trusted sources for username matching in settings. This feature allows to define and manage the trusted data sources for username matching from the configuration.


sduenas (approved):
LGTM

PR #935, opened 2024-10-03
Correct arrow direction when merging recommendations

### LINKED ISSUE #934: "Review recommendations" has inverted item order
In the web UI, "Review recommendations: Is this the same individual?" displays `foo → bar`.

This implies that `foo` will be merged into `bar`, and that `bar` will survive.

But after clicking the "Merge" button, the resulting entity is named `foo` instead of `bar`.

The easiest fix would probably be changing https://github.com/chaoss/grimoirelab-sortinghat/blob/master/ui/src/components/Recommendations.vue#L100 from `mdi-arrow-right` to `mdi-arrow-left`.

aklapper:
Currently shows `foo → bar` but direction is inverted as the resulting entity will be named `foo` instead of `bar`.

Closes #934

### DIFF
--- a/releases/unreleased/wrong-arrong-direction-when-merging-recommendations.yml
+++ b/releases/unreleased/wrong-arrong-direction-when-merging-recommendations.yml
@@ -1,0 +1,11 @@
+---
+title: Wrong arrong direction when merging recommendations
+category: fixed
+author: Andre Klapper
+issue: 934
+notes: >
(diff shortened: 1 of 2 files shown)


sduenas (approved):
LGTM

sduenas:
Thanks for the patch @aklapper . I updated your PR to include the missing changelog file.

PR #939, opened 2024-11-06
[ui] Skip recommendation

### LINKED ISSUE #937: Review Recommendations: Skip option
The recommendations review loops over all recommendations in sequence. When the user gets a doubtful match she has 3 choices:
- Confirm the match (merge)
- Dismiss (forever)
- Stop the review. In this case, the pending recommendations are stuck until she decides to reset (remove all pending recommendations)

A 'skip/go to next recommendation' option would allow the user to solve all clear recommendations before stalling or resetting.

evamillan:
This PR adds a button on the "Review recommendations" modal to skip the current recommendation and load the next one.
Fixes #937.

### DIFF
--- a/releases/unreleased/skip-recommendations.yml
+++ b/releases/unreleased/skip-recommendations.yml
@@ -1,0 +1,9 @@
+---
+title: Skip recommendations
+category: added
+author: Eva Millán <evamillan@bitergia.com>
+issue: 937
+notes: >
+  Users can now skip a recommendation and review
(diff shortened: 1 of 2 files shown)


sduenas (changes requested):
The PR looks ok but when there's only one recommendation or it's the last recommendation, the `Skip` button doesn't show. In those cases, the button should be there but once is clicked, the modal hides.

sduenas (approved):
LGTM

PR #942, opened 2024-11-15
Run tests using Testcontainers

jjmerchante:
This PR simplifies the process of running SortingHat tests by using Testcontainers. With Testcontainers, we can now easily set up containerized services, creating a consistent testing environment and reducing setup complexity.


sduenas (approved):
This looks great!

PR #943, opened 2024-11-20
Sort organizations by enrollments

### LINKED ISSUE #940: Show in the companies table the ones with enrollments
If we think of SortingHat as the UI to polish up a standalone data set, there are several ways to improve the table on the right with the list of organizations/companies. Right now, that table is not useful, as it always shows the same companies and in most cases with 0 enrollments. What would be interesting is to see the companies with enrollments.

imagen

In a conversation with @evamillan, she had a simple idea that can make it useful. The idea is to sort the companies by the number of enrollments. I would do this by default.

evamillan:
This PR adds the option to sort the list of organizations by number of enrollments. On the UI, the table columns are made actionable to sort by those fields and the default is enrollments by descending order, to show the organizations with the most enrollments first.
Fixes #940

### DIFF
--- a/releases/unreleased/order-organizations-by-number-of-enrollments.yml
+++ b/releases/unreleased/order-organizations-by-number-of-enrollments.yml
@@ -1,0 +1,9 @@
+---
+title: Sort organizations by number of enrollments
+category: added
+author: Eva Millán <evamillan@bitergia.com>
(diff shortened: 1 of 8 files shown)


sduenas (approved):
LGTM

PR #949, opened 2025-01-15
Update version of django-cors-headers

jjmerchante:
This commit updates the version of the `django-cors-headers` package to the latest version available.

This is needed to ensure compatibility with `grimoirelab-core`, which uses a newer version.


sduenas (approved):
LGTM

PR #970, opened 2025-05-14
[ui] Fix dates different from input

### LINKED ISSUE #966: Dates get saved differently from how they were input
I have several instances where I enter a date for an Organizational affiliation and the date that is shown after it saved is different, usually a few days earlier.

I wonder if this is a timezone issue.

How to reproduce:
- Edit an affiliation and put a date like 2022-10-01
- Save, and the profile shows 2022-09-28

evamillan:
This PR fixes the issue with the affiliation input dates being different from the saved value in some timezones.

Fixes #966.


sduenas (approved):
LGTM

PR #971, opened 2025-05-14
Allow pasting LinkedIn URLs

### LINKED ISSUE #967: Allow pasting LinkedIn URLs
When I add LinkedIn to a profile, I find it easiest to copy the LinkedIn URL. However, SortingHat only wants a specific part of the URL and it breaks the link if I post the full URL.

I suggest that SortingHat allows pasting the full URL and does the cleanup for the user.

For example, a JavaScript RegEx could clean the input to remove the rest of the URL. Also, there should be a server-side check that a full URL is not passed.

evamillan:
This PR allows users to enter a URL when they add a LinkedIn profile. The input value is cleaned and only the slug of the URL is saved. A server-side validation is also added to check that a URL was not passed.

Fixes #967.

### DIFF
--- a/releases/unreleased/allow-pasting-linkedin-url.yml
+++ b/releases/unreleased/allow-pasting-linkedin-url.yml
@@ -1,0 +1,8 @@
+---
+title: Allow pasting LinkedIn URL
+category: added
+author: Eva Millán <evamillan@bitergia.com>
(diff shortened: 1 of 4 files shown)


sduenas (approved):
LGTM

Issue #988, opened 2025-07-28
API response is slow due to changelog information


sduenas:
When I was working with the UI to get info from some individuals, I noticed the responses were really slow. Checking the POST requests, the responses took between 16 and 20 seconds to generate.

Having a look to the GraphQL queries, I realized they are getting information from the transactions/operations tables looking for the UUID/mk of the individual. The problem is the queries use the column `target` of the table `operations` to search for the given individual. This column has a `VARCHAR` and doesn't have an index. Therefore, the search needs to check all the rows to get the data.

To fix this problem, we have to add an index on the column `target`.

PR #989, opened 2025-08-01
Performance improvements

jjmerchante:
- Added an index on the target field in the `Operation` table to improve the performance of changelog queries. Previously, these queries used the entire table to find matching rows, resulting in slow response times.

- Added a `totalEnrollments` field to the Organization query to return just the count of enrollments. This avoids fetching all enrollments for each organization, reducing database load and speeding up table rendering.

Fixes https://github.com/chaoss/grimoirelab-sortinghat/issues/988


sduenas (approved):
LGTM

Issue #994, opened 2025-08-25
Affiliate job overwrites affiliation dates


evamillan:
Running an "affiliate" job overwrites the existing affiliation dates.

Before affiliate:

<img width="1196" height="960" alt="Image" src="https://github.com/user-attachments/assets/e4994177-2bea-4f79-9300-6a68468482ec" />

After affiliate:

<img width="1207" height="1011" alt="Image" src="https://github.com/user-attachments/assets/37e4395c-f2d7-4cd8-b637-4cff3d60b620" />

PR #996, opened 2025-08-28
Correct affiliate process to prevent overwriting

jjmerchante:
This change fixes a bug in the affiliate process. Previously, the process would overwrite existing affiliation dates, leading to incorrect results. Now, it only includes individuals who are newly affiliated, ensuring accurate data retention.

Fixes https://github.com/chaoss/grimoirelab-sortinghat/issues/994


sduenas (approved):
LGTM

sduenas (approved):

PR #1000, opened 2025-09-08
[ui] Improve merge recommendations

evamillan:
This PR changes the merge recommendations modal to show more than one at a time. The recommendations are ordered by individual so they can be grouped on the UI to manage them in batches. Users can select which recommendations to apply or dismiss using checkboxes. The list of recommendations is also paginated to improve the navigation.

recs-grouped

### DIFF
--- a/releases/unreleased/improved-ui-for-merge-recommendations.yml
+++ b/releases/unreleased/improved-ui-for-merge-recommendations.yml
@@ -1,0 +1,10 @@
+---
+title: Improved UI for merge recommendations
+category: added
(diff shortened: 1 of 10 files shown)


evamillan:
I updated the PR, keeping one set of recommendations per page like before, but ordering by number of recommendations and showing all of the individual's recommendations at once.

recs-modal-ordered

evamillan:
I replaced the checkboxes with buttons to make the actions being applied clearer.

buttons

